PR #21, opened 2020-12-14
Replace ```ignore with ```text

lilyball:
Using the `ignore` marker makes `cargo test` think this is rust code that should be ignored, which means it prints an "ignored" line in the test output. It also makes `cargo doc --document-private-items` attempt to colorize it as rust code. Marking this as `text` instead fixes both issues.


zkat:
aaaaa that was the magic word! Thank you!

zkat (approved):

PR #33, opened 2020-12-16
Tighten up slashdash definition

### LINKED ISSUE #31: Slashdash comments are... wrong
You get some really weird constructs with the current spec. @lilyball suggested a way to fix it, though! See #29

lilyball:
Previously slashdash was allowed anywhere `ws` was, and could be followed by any legal slashdashable token regardless of position. This caused some ambiguous parsing and enabled constructions we really don't want.

The new version simply allows a slashdash to preceed each commentable token, thus restricting its use to locations where the token is otherwise valid.

Fixes #31.


zkat (approved):
this is awesome! Thank you so much for this! :)

PR #34, opened 2020-12-16
bring back bom-as-ws

zkat:
how's this, @lilyball ?

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -194,7 +194,7 @@ linespace := newline | ws | single-line-comment
 
 newline := `000D` | `000A` | `000D` `000A` | `0085` | `000C` | `2028` | `2029`
 
-ws := bom |unicode-space | multi-line-comment
+ws := bom | unicode-space | multi-line-comment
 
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -198,7 +194,9 @@ linespace := newline | ws | single-line-comment
 
 newline := `000D` | `000A` | `000D` `000A` | `0085` | `000C` | `2028` | `2029`
 
-ws := unicode-space | multi-line-comment
+ws := bom |unicode-space | multi-line-comment
+
+bom := `FFEF`
```
lilyball, line 199: I know we’re already using it elsewhere, but I’m not terribly fond of backticks to denote unicode scalars. I’d rather we use `U+FFEF` instead, which is the same number of characters anyway.

lilyball (commented):
LGTM except for the missing space.

SPEC.md
zkat: Yeah let's fix that (I'm going to bed for the night so tomorrow?

PR #44, opened 2020-12-22
Fixed escape table formatting

hclarke:
table wasn't displaying right on github.

this is a neat language :)


zkat (approved):
TYSM! I can never get markdown tables right haha

PR #57, opened 2020-12-27
identifier and string tweaks

hclarke:
this does a few things with identifiers and strings:

- removes `"` from the set of identifier characters, to prevent ambiguity with string
- adds `bare-identifier` and `escaped-string` non-terminals (i think less mixing of terminal/non-terminal is easier to read?)
- moves string ahead of bare-identifier in identifier, for easier implementation with ordered-choice parsers
- moves raw-string into string, so you can have raw-string identifiers

(i'd be happy to break this up, if parts need more discussion)


zkat (approved):
These changes all look good to me! :) Thank you!

PR #60, opened 2020-12-28
Fill in string escapes table

larsgw:


larsgw:
I'm not entirely sure why this was not filled in yet, if this table wasn't final then I guess just close the PR.

zkat (approved):
Ah! Thank you!

It wasn't filled in because I was working on other stuff and forgot all about this. Thanks!

larsgw:
So backslashes followed by non-escape characters are just invalid right? Not that it returns the letter/codepoint it escapes? The only possible negative effect I can think of is that if people copy text with escaped single quotes into KDL they have to remove the escapes, but that's hardly even a negative effect.

zkat:
correct

Issue #46, opened 2020-12-23
Add a logo for KDL


obeid:
KDL does not _need_ a logo, but as an early fan, I _want_ it to have a logo.

timmybytes:
I did a quick little mock-up of a logo, pretty heavily emphasizing the "cuddly" feel of the KDL name in case you're interested @zkat? It has a transparent background, so it works with light and dark backgrounds.

![KDL_Logo_Draft](https://user-images.githubusercontent.com/67444371/103307922-0ae40500-49c6-11eb-8506-03c7e7667329.png)

If you like it I can submit a PR, or if you have suggestions I can work on different versions!

zkat:
Um, I actually _love_ this? It is extremely my kind of vibe. I'm wondering what others think, too? Because in my mind, we'd just adopt this right now hahah.

obeid:
I made the issue so I can design the logo, but I wouldn't be mad if this gets used. I'm feeling it too. Great job @timmybytes

timmybytes:
Oh sorry, @obeid! I didn't mean to step on any toes, I just geek out on designing these kinds of things. But I'm so glad you and @zkat like it! I'll tweak a couple things to get the design a little more polished, and make some different versions for various use cases (repo sharing card, svg, logo without tagline, etc.) if that's okay?

Rupertofly:
Ah, the slippery slope of branding, if we use it, could you let us know the colours and font used?

timmybytes:
The font is [Nunito](https://fonts.google.com/specimen/Nunito), an Open Font licensed Google Font. I made six different sizing variations (though I'm having issues with exporting to SVG), and included a simple styles sheet (shown here as a PNG since I can't attach the PDF). While the "Medium" and "Large" sized canvases are the same size, the medium has more empty space for breathing room. No worries at all if you all end up going with something different. If anything, it was fun to practice with!

![Styles](https://user-images.githubusercontent.com/67444371/103321690-bd7c8d80-49ef-11eb-9451-aa2800d8d4d4.png)
![KDL_Logo_Only_Large_1280x640](https://user-images.githubusercontent.com/67444371/103321700-c4a39b80-49ef-11eb-8791-721df6aaf7da.png)
Large 1280x640 Logo

![KDL_Logo_Only_Medium_1280x640](https://user-images.githubusercontent.com/67444371/103321703-c9684f80-49ef-11eb-88b8-f309d8cd5fad.png)
Medium 1280x640 Logo

[...]

PR #61, opened 2021-01-01
(style): Add KDL logo and favicon files

timmybytes:
Added `graphics` directory with various KDL logo sizes, favicons, and style guide reference based on discussion in #46. Left README unmodified so maintainers can best decide how/where to implement logos.


zkat (approved):
Thank you so much!

PR #71, opened 2021-01-08
Correct description of the raw string format

hkolbeck:
By my understanding of the grammar, raw strings are formatted `r###"the\nstring"###`, with quotes surrounding the inner string, rather than `r"###the\nstring"###`, with the start and end tokens equal. This changes the description to match that definition.


zkat (approved):
Oh! Nice catch

PR #72, opened 2021-01-10
Possible typo in "Escapes" table

jellymann:
The maximum unicode code point is `10FFFF` but in the escapes table it states code points can only go up to `10FFF`. I think this is an error, since most emojis are well above that number, and other languages allow escapes up to `10FFFF`


zkat (approved):
oh I thought I'd fixed this! Or maybe I fixed it elsewhere but not here. Thank you!

PR #73, opened 2021-01-10
Add forwardlash to escapes table

jellymann:
I noticed this was a valid escape char in the rust implementation. Not sure if this was intended.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -186,6 +186,7 @@ interpreted as described in the following table:
 | Carriage Return               | `\r`   | `U+000D` |
 | Character Tabulation (Tab)    | `\t`   | `U+0009` |
 | Reverse Solidus (Backslash)   | `\\`   | `U+005C` |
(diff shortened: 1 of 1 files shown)


zkat (approved):
hahaha. This was definitiely an oversight. Thank you!

hkolbeck-streem:
I was wondering about that 😅 in my testing I was surprised to find that a/**/ was a valid bare id depending on how you read the spec.

zkat:
Ugh. This makes me want to remove / and * as valid identifier characters :/

hkolbeck-streem:
'-' is also fraught, "-10"/+10 are valid bare identifiers, if I'm not mistaken 😬 though at least my impl will only treat them such in the node identifier position.

There at least you could just add it to the list of illegal chars at the start of a bare id.

zkat:
Would you file an issue so we can take a closer look at these?

hkolbeck-streem:
Will do.

Issue #85, opened 2021-04-11
Support signs for non-decimal numbers


zkat:
Right now, you can only use signs with decimal numbers. Let's extend that to include the other formats.

hkolbeck:
I am swamped with other projects if someone would like to take on handling this for [kdl4j](https://github.com/hkolbeck/kdl4j)

PR #87, opened 2021-04-12
Added optional sign to all number syntaxes

daeken:
This is a spec update to reflect issue #85. Updates the BNF and spec wording regarding signs on non-decimal numbers.


larsgw:
LGTM

PR #92, opened 2021-07-18
Add JiK spec

tabatkins:
Did what I could from <https://twitter.com/zkat__/status/1416833847908392960>, but it doesn't look like the website is built from this repo, so I couldn't add to the website's faq.

Also, decided to commit to JSON's names, and used "object" as the node name rather than "dict". Could alternately go with "list" and "dict" to go whole-hog with slightly-shorter Pythonish names; it just felt weird to have one JSish name and one Pythonish name.

### DIFF
--- a/JSON-IN-KDL.md
+++ b/JSON-IN-KDL.md
@@ -1,0 +1,91 @@
+JSON-in-KDL (JiK)
+=================
+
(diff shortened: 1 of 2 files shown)


zkat (approved):

zkat:
aaaa this is so great. Thank you so much! And yeah, the website's actually built from https://github.com/kdl-org/kdl-org.github.io

PR #93, opened 2021-07-19
Switch literal nodes to `-`

tabatkins:
`-` is slightly easier to type than `_` (no Shift, at least on american keyboards), and it's consistent with the "dummy" node name I'm planning to use for text nodes in XML-in-KDL.

(`_` is a valid XML element name, so I can't use that in XiK.)


tabatkins:
Just realized that this is literally a breaking change, my lack of commitment to semver is showing.

PR #95, opened 2021-08-03
First pass at XiK

tabatkins:
Similarly to my last JiK spec, XiK is a lossless encoding of XML, for interoperating with XML toolchains.

Luckily, KDL already resembles XML quite closely, so for the most part it's just a slightly more convenient XML syntax.

I've also adjusted the `website.kdl` doc to conform to this; the only change was to fix the spelling of the doctype to `!doctype`, and change the text nodes to `-`.

### DIFF
--- a/XML-IN-KDL.md
+++ b/XML-IN-KDL.md
@@ -13,7 +13,7 @@ XML has several types of nodes, corresponding to certain KDL constructs:
 
 * Elements, which have an element name, zero or more attribute, and zero or more children. These are encoded directly as KDL nodes, using the nodename, properties, and children nodes.
 * Raw text. In "pure" XML dialects, where raw text only appears as the sole child of an element (never mixed with other elements as siblings), this is generally encoded as a final string argument in a KDL node; in "mixed" XML dialects, it can be encoded as a special KDL node with the name `-`.
-* Comments. These are encoded as KDL nodes with the name `!`.
+* Comments are encoded as KDL block comments. (Or as an actual node type, for some use-cases.)
 * Processing Instructions. These are encoded similarly to elements if their contents are sufficiently structured, with a `?` in front of their node name. If they're not sufficiently structured, their contents are just strings.
 * Doctypes. These are encoded like unstructured PIs, just with the node name `!doctype`.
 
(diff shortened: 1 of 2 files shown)


XML-IN-KDL.md
```diff
@@ -0,0 +1,79 @@
...
+
+If the element contains mixed text and element children, the text can be encoded as a KDL node with the name `-` with a single string unnamed argument. For example, the XML `<span>some <b>bold</b> text</span>` can be encoded as `span { - "some "; b "bold"; - " text" }`.
+
+An element that contains only text *is allowed to* encode it as `-` children. For example, `<span>foo</span>` *may* be encoded as `span { - "foo" }` instead of `span "foo"`. However, an element cannot mix the "final string attribute" with child nodes; `span "foo" { b "bar" }` is an **invalid** encoding of `<span>foo<b>bar</b></span>`. (It must be encoded as `span { - "foo"; b "bar" }`.)
+
+CDATA sections are not preserved in this encoding, as they are merely a source convenience so you don't have to escape a bunch of characters. They are encoded as normal textual contents would be.
+
+-----
+
+Comments are encoded as a special node name `!`, with a single unnamed string argument containing the comment's value (everything between the `<!--` and the `-->`). For example, `<!-- comment! -->` is encoded as the node `! " comment! "`.
```
zkat, line 46: maybe I don't understand XML enough, but what's the reason for using a node for comments, instead of KDL's existing comment facilities?

tabatkins: I'm *presuming* that XML comments should round-trip (but KDL comments shouldn't, at least not necessarily).

I could be convinced otherwise.

tabatkins: Maybe just have a note to only use this when authoring if you *specifically* want the comment to reflect into the XML document; otherwise stick with KDL comments?

zkat: I like that idea!

Although I'm also not opposed to saying you can/should use `/* */` comments (only) if you want them to round-trip, and all other KDL comments can be ignored as you said (or even disallowed. `/-` is pretty weird when it comes to comment formats)

tabatkins: Okay, done. I said that they're encoded as multiline comments, but left `!` nodes as an alternative option specifically if you're using a KDL toolchain that discards comments and need them reflected into the XML.

zkat (approved):

zkat:
Thanks for this! This is really great! I hope it was fun to write :)

Lucretiel:
Just a heads up that there's actually a conflict here with the recent numbers PR, in that this spec calls for mixed style XML text child nodes to be encoded with the name `-`, which is not longer allowed. This could potentially be made compatible with the restriction on identifier names by loosening them slightly such that identifiers are prohibited from starting with `+` or `-` only in the case where the next character is a digit (which covers both decimal numbers and the `0` prefix of other numbers).

tabatkins:
Yeah, personally I'd like the "don't let idents look like numbers" specifically disallow `^[+-]\d`, not just any starting +-.

Or just allowing `-` all on its own, that's def not confusable with numbers.

zkat:
@tabatkins i'm fine with that change. would you like to PR it? :)

PR #99, opened 2021-08-24
identifiers should never look like numbers

zkat:
This does mean that -foo is now illegal, when it wasn't before, but I think it's for the best?

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -311,7 +311,7 @@ node-space := ws* escline ws* | ws+
 node-terminator := single-line-comment | newline | ';' | eof
 
 identifier := string | bare-identifier
-bare-identifier := (identifier-char - digit) identifier-char*
+bare-identifier := (identifier-char - digit - sign) identifier-char*
 identifier-char := unicode - linespace - [\{}<>;[]=,"]
 prop := identifier '=' value
 value := string | number | boolean | 'null'


Lucretiel:
So first of all this is breaking change, so obviously be sure to document is as such if it merge.

I'm actually going to push for that only the first letter should be disallowed from a sign, because I think we explicitly *do* want to allow identifiers like `kebab-case-word` and `kdl-rs`

zkat:
(an advantage of a language with basically no users is that breaking changes are basically free, and other existing impls can just make sure they conform before saying they support 1.0.0)

I really need to write up a test suite of valid and invalid KDL documents, though.

Lucretiel:
> isn't that what I changed the spec to? This only applies to the _first_ character, not the other identifier chars. So `-foo` is illegal but `foo-bar` is fine.

Oh, you're right! I sort of thought that there was a "bare-identifier-start" and that's what my eye was looking for. Looks great then!

> I think once you're done with your parser,

It's funny you should mention this because 🤣😭 https://github.com/Lucretiel/kaydle/blob/6cf8a99e830356bed94275a67e01ff5fe4979484/src/primitives/string.rs#L143-L160 I'll go back and fix it later I promise!

Lucretiel:
> I really need to write up a test suite of valid and invalid KDL documents, though.

I was just gonna ask about this! I can help make this after kaydle is ready but yeah I was looking earlier for a collection of documents especially to test all the edge cases with whitespace and component boundaries and so on

zkat:
https://github.com/hkolbeck/kdl4j/tree/trunk/src/test/resources/test_cases/input maybe @hkolbeck's test suite is a good starting point?

hkolbeck:
@Lucretiel I'm happy to provide any help I can on moving my test suite into a more widely accessible place. The examples @zkat linked to are matched to expected outputs in a sibling-directory, if no expected output is present then parsing should fail. 

The code for running these tests can be found here: https://github.com/hkolbeck/kdl4j/blob/trunk/src/test/java/dev/hbeck/kdl/TestRoundTrip.java

It's worth noting that I have been unable to find time to keep KDL4j up to the current language specification, so some of these tests may be dated. If either of you knows a java-type-person who might be interested in helping out with that update, it would be much appreciated.

Issue #114, opened 2021-09-01
"keywords" are valid identifiers


zkat:
### Discussed in https://github.com/kdl-org/kdl/discussions/49

<div type='discussions-op-text'>

<sup>Originally posted by **hclarke** December 24, 2020</sup>
it looks like `true`, `false`, and `null` are valid identifiers

that doesn't really conflict with anything, but it seems a bit weird?

options i see:
- explicitly disallow that
- allow any value as an identifier (not just stringy ones)
- leave it as-is</div>

zkat:
I've decided not to do this. I think the "surprise" of parsing failing because you didn't know KDL had keywords (which is a reasonable assumption to make) is more damaging than the potential confusion when reading code that uses these.

zkat:
I changed my mind again. Twitter also set me straight.

PR #117, opened 2021-09-01
prevent the three "keywords" from being used as bare identifiers

zkat:
Fixes: https://github.com/kdl-org/kdl/issues/114


zkat:
I've decided not to do this. I think the "surprise" of parsing failing because you didn't know KDL had keywords (which is a reasonable assumption to make) is more damaging than the potential confusion when reading code that uses these.

zkat:
changed my mind :P

Issue #113, opened 2021-09-01
Proposed extensions to the schema specification


zkat:
### Discussed in https://github.com/kdl-org/kdl/discussions/107

<div type='discussions-op-text'>

<sup>Originally posted by **larsgw** August 31, 2021</sup>
  1. A top-level `metadata` (or `info`) node. Authors, title/description, URLs (documentation, persistent self-URL for linked data), publication dates and version information.
  2. A top-level `definitions` node. In the example schema, the props `description`, `ref` and `id` are shared among various nodes throughout the documentation. To remove the duplication one could give the first definition an `id` and `ref` it in other places but the first definition is not necessarily the most natural place to put it. A `definitions` node could house `node`, `prop`, `value` and `children` nodes.

(Edit: I see there are `description` props already on `document`, but I think more metadata is still helpful. The same goes for `schema-url` if it's a reference to the current schema; if it's a reference to the schema schema, like an XML namespace, then a top-level `schema` node to contain `metadata`, `definitions` and `document` might be good too.)</div>

PR #120, opened 2021-09-01
Add schema specs for info and definitions

larsgw:
Fix https://github.com/kdl-org/kdl/issues/113

### DIFF
--- a/SCHEMA-SPEC.md
+++ b/SCHEMA-SPEC.md
@@ -29,9 +29,124 @@ None.
 
 #### Children
 
+* [`info`](#info-node) - one info node for that describes the schema itself.
 * [`node`](#node-node) - zero or more toplevel nodes for the KDL document this schema describes.
+* [`definitions`](#definitions-node) (optional): Definitions of nodes, values, props, and children block to reference in the toplevel nodes.
 * `node-names` (optional): [Validations](#validation-nodes) to apply to the _names_ of child nodes.
 * `other-nodes-allowed` (optional): Whether to allow nodes other than the ones explicitly listed here. Defaults to `false`.
+
+### `info` node
+
+The `info` node describes the schema itself.
+
+#### Values
+
+None.
+
+#### Properties
+
+None.
+
+#### Children
+
+* [`title`](#title-node): one or more titles
+* [`description`](#description-node) (optional): zero or more descriptions
+* [`author`](#author-and-contributor-nodes): one or more authors
+* [`contributor`](#author-and-contributor-nodes) (optional): zero or more contributors
+* [`link`](#link-node) (optional): zero or more URLs
+* [`license`](#license-node) (optional): zero or more licenses
+* [`published`](#published-and-modified-nodes) (optional): a publication date
+* [`modified`](#published-and-modified-nodes) (optional): a modification date
+* [`version`](#version-node) (optional): a (semver) version number
+
+### `title` node
+
+The title of the schema or the format it describes.
+
+#### Values
+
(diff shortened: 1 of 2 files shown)


SCHEMA-SPEC.md
```diff
@@ -24,15 +24,129 @@ None.
...
+
+None.
+
+#### Properties
+
+None.
+
+#### Children
+
+* [`title`](#title-node): one or more titles
```
zkat, line 51: Can we make all children optional here? I think requiring only specific child nodes here is a bit awkward?

SCHEMA-SPEC.md
```diff
@@ -24,15 +24,129 @@ None.
...
+* [`description`](#description-node) (optional): zero or more descriptions
+* [`author`](#author-and-contributor-nodes): one or more authors
+* [`contributor`](#author-and-contributor-nodes) (optional): zero or more contributors
+* [`link`](#link-node) (optional): zero or more URLs
+* [`license`](#license-node) (optional): zero or more licenses
+* [`published`](#published-and-modified-nodes) (optional): a publication date
+* [`modified`](#published-and-modified-nodes) (optional): a modification date
+* [`version`](#version-node) (optional): a (semver) version number
```
zkat, line 59: Could you get rid of the () around semver and link the word to semver.org?

SCHEMA-SPEC.md
```diff
@@ -24,15 +24,129 @@ None.
...
+
+* `lang` (optional): An IETF BCP 47 language tag
+
+### `description` node
+
+A description of the schema or the format it describes.
+
+#### Values
+
+* Description
```
zkat, line 79: The definition for `info` says this can be "zero or more". I think I'd rather this be min1max1 instead of several blocks of text? Is there a use case for that?

SCHEMA-SPEC.md
```diff
@@ -24,15 +24,129 @@ None.
...
+
+Author(s) of the schema.
+
+#### Values
+
+* Author name
+
+#### Properties
+
+* `orcid` (optional): The [ORCID](https://orcid.org/) of the author.
```
zkat, line 95: I hadn't heard of this! This is really cool!

SCHEMA-SPEC.md
```diff
@@ -24,15 +24,129 @@ None.
...
+
+* [`link`](#link-node) (optional): zero or more URLs
+
+### `link` node
+
+Links to the schema itself, and to sources about the schema.
+
+#### Values
+
+* URL - A URL that the link points to
```
zkat, line 107: Can this be `type "uri" "iri"`?

zkat (commented):
Aaaaa this looks genuinely so good! Thank you! Just a few comments that really are mostly musings and I think it's ok to dismiss most of them if you dislike/disagree

SCHEMA-SPEC.md
larsgw: Multiple languages is what I had in mind. But there's no `max` validator that requires at most one for each value of a certain property.

SCHEMA-SPEC.md
larsgw: Of course, I hadn't thought of them.

SCHEMA-SPEC.md
larsgw: That's fair, I was coming from a more bibliographical perspective but that's not necessary here.

zkat (approved):
This LGTM! Thanks a bunch!

Issue #121, opened 2021-09-01
Deviations from the spec in the kdl4j test suite


larsgw:
When I implemented the kdl4j test suite in kdljs I came across a few deviations from the spec. Since some of them are sensible I wanted to discuss them here.

### Multiline comments in node-space

The spec does currently not allow for the following, neither in the grammar or the spec text:

```kdl
node /* comment */ "arg"
```

### Single-line comments at EOF

The spec currently requires a single-line comment to be ended by a newline, which makes the following invalid (without a trailing newline):

```kdl
// hi
```

### Esclines between `/-` and props/values/children blocks

The spec currently only allows for pure whitespace between `/-` and the thing it is commenting out. Therefore, the following is invalid (I would definitely agree with the spec here, I just added it for completeness):

```kdl
node /-    \
    "arg"
```

### Escline outside node-space

Esclines can currently only be in node-space, so the following is invalid (again, I agree):

```kdl
node1
  \// hey
   node2
```

### Large numbers

The question also arose on how to deal with large numbers. Other than considering BigInt and BigDecimal, parsing *large* integers as `Infinity` seems relatively fine to me. However, a problem arises when `Infinity` should be formatted. Should `Infinity`, `-Infinity` and `NaN` be allowed for full float compatibility?

zkat:
> Multiline comments in node-space
> Single-line comments at EOF

These all smell like spec bugs to me. What do you think? Should we fix this in spec, or consider them quirks?

> Esclines between /- and props/values/children blocks

I'm on the fence about this one. I feel like we should be treating esclines as if they were single-line spaces? Maybe we should fix the spec here, even though this is a weird corner case.

> Escline outside node-space

That doesn't look like a linespace to me. Which makes me realize that \/*foo*/ *should* be valid because multi-line comments _are_ allowed as part of linespace???

> Large numbers

I've been thinking about this one a bit. What do you think of saying "All numbers that, when interpreted as 64-bit IEEE 754 floating point numbers MUST be losslessly accepted by KDL parsers. Larger (or "smaller", if negative) numbers MAY be supported by individual KDL implementations, but are not guaranteed to be portable."?

hkolbeck:
Apologies for the misaligned tests, but I'm glad they're exposing some worthwhile corner cases.

On the large number question, I think it's worth discussing the full gap between a float64 and a BigDecimal, specifically including a mention of underflow.

zkat:
[...]

Issue #106, opened 2021-08-30
Extending primitives


tabatkins:
#89 requests a new primitive type, the "symbol", which is similar to but distinguishable from strings so they can unambiguously distinguish between "constant values" (strings) and "variables" (symbols) in a value list.

There have been requests for other additional primitives as well, such as dates, or binary blobs. Right now either of these have to be encoded as strings, and require domain-specific knowledge to tell apart from "ordinary" strings; this also means you can't use the primitive's type to distinguish things and must instead rely on named values or a required position.

The only other way around these issues is to move this data down into the child list, encoding it as nodes: `symbol "foo"`, `date "1970-01-01"`, `base64 "..."`, etc. But then you run into the issue mentioned in #105, where you can no longer identify these nodes with names like named values, nor can you mingle them with other primitives; they're segregated into the child list on their own.

I agree with your stance so far on keeping the set of primitives small and simple, but believe a simple extension mechanism would allow most of these scenarios to be addressed with minimal pain.

----

Proposal: Add "tagged strings" as a new primitive type, written as `tag@"string"`. The @ separates the tag from the string syntax (so it's unambiguous with the `r` string prefix), and all string syntaxes are allowed.

No meaning is given to these; the tag is just an opaque identifier carried with the string, letting the user identify the data as having some specific additional meaning. Microsyntaxes within the string are still the user's responsibility; there's no further exposed structure (in particular, no nesting).

(I *think* the @ doesn't interfere with ident syntax, which allows `@` anywhere in it; you could have a tag named "foo@" by writing `foo@@"string"`.)

So dates could then be easily distinguished as `date@"1970-01-01"`, base-64-encoded data as `bin@"..."`, etc. I think the empty tag should be allowed as well, like `@"foo"`, to allow for easy encoding of trivial cases like the symbol/string distinction from #89.

zkat:
I feel as though something like this might be better served by declaring desired interpretations in a Schema? (see #104) What's the actual value of inlining this kind of metadata? Is it a dynamic languages thing? Stuff like serde has no issue being told what a particular value is supposed to be, and I guess that's because the type defs are a schema of sorts...

zkat:
It's late and I can't stop thinking about this and I'm now leaning towards type hints being a good and useful thing to have inline? I guess I'll sleep on it. I like keeping things simple but this would be a small addition, and I don't think there's any harm in having type *hints* if we're already expecting strings to be interpreted in different ways based on schema/context?

[...]

PR #122, opened 2021-09-01
add a note about representations

zkat:
Ref: https://github.com/kdl-org/kdl/issues/121

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -243,11 +243,10 @@ It's up to individual implementations to determine how to represent KDL
 numbers.
 
 Implementations MUST support number values representable as IEEE754 floating
-point numbers. That is, all implementations must support at least the numbers
-between `-(2^53 - 1)` and `2^53 - 1`. Larger or smaller values MAY error, or
-be represented as `Infinity`, or `NaN`, or any other value of the
-implementation's choosing. If you want to reliably represent numbers outside
-of this range, use the `string` type to represent them.
+point numbers. Larger or smaller values MAY error, or be represented as
+`Infinity`, or `NaN`, or any other value of the implementation's choosing. If
+you want to reliably represent numbers outside of this range, use the `string`
+type to represent them.
 
 There are four syntaxes for Numbers: Decimal, Hexadecimal, Octal, and Binary.
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -242,6 +242,13 @@ distinction in KDL between real numbers, integers, and floating point numbers.
 It's up to individual implementations to determine how to represent KDL
 numbers.
 
+Implementations MUST support number values representable as IEEE754 floating
+point numbers. That is, all implementations must support at least the numbers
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -242,6 +242,13 @@ distinction in KDL between real numbers, integers, and floating point numbers.
 It's up to individual implementations to determine how to represent KDL
 numbers.
 
+Implementations MUST support number values representable as IEEE754 floating
+point numbers. That is, all implementations must support at least the numbers
+between `-(2^53 - 1)` and `2^53 - 1`. Larger or smaller values MAY error, or
```
hkolbeck, line 247: nit: There are lots of numbers between those two that are not storable in a 64 bit float. Maybe just remove the "That is..." sentence and let implementors refer to IEEE754 directly if needed?

zkat: ok!

larsgw:
Is there a canonical representation of IEEE754 `Infinity` (and `NaN` and `-Infinity`) in KDL? Making them `null` or removing them entirely could break the resulting document, but there is no `Infinity` "keyword" in KDL either. Should a writer instead return an exception when an `Infinity` is passed?

zkat:
There is not, as far as I know.

Precedent-wise: YAML supports `.inf`, `-.inf`, and `NaN`. JSON has no Infinity or NaN. It just turns things into `null` afaict.

This PR originally had me adding `infinity` and `nan` as keywords but honestly I think that's a bad idea and the answer should be "use a string if you want really big numbers".

As to your question: It's entirely up to you what happens if someone passes in Infinity. It is canonically nasal demons territory. You could throw an error and tell people there's a number overflow/underflow and to use strings instead?

zkat:
@hkolbeck @larsgw so I've been thinking about this, and how "convenient" the fact that no format for numbers was actually specified in KDL proper. How do y'all feel about not landing this, and just say "numbers are just syntax. Good luck, have fun", and tell users instead to use KDL Schema (or tagging them inline, per #106)?

So if you wanted to say "this is DEFINITELY a u8", you'd do `<u8>1.2345`? (or whatever syntax we come up with for that, or specify it in the KDL Schema instead)? I feel like I'd rather do that and keep KDL "simple"?

larsgw:
> How do y'all feel about not landing this, and just say "numbers are just syntax. Good luck, have fun", and tell users instead to use KDL Schema (or tagging them inline, per #106)?
>
> So if you wanted to say "this is DEFINITELY a u8", you'd do <u8>1.2345? (or whatever syntax we come up with for that, or specify it in the KDL Schema instead)? I feel like I'd rather do that and keep KDL "simple"?

That's fine with me. I don't really like the tagging myself but specifying in the schema is fine.

> As to your question: It's entirely up to you what happens if someone passes in Infinity. It is canonically nasal demons territory. You could throw an error and tell people there's a number overflow/underflow and to use strings instead?

Yeah, I think I prefer errors to silently changing it to JSON.

zkat:
closing this, then

PR #123, opened 2021-09-02
add type annotation info to spec

zkat:
Fixes: https://github.com/kdl-org/kdl/issues/106

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -189,6 +189,78 @@ A value is either: a [String](#string), a [Raw String](#raw-string), a
 Values _MUST_ be either [Arguments](#argument) or values of
 [Properties](#property).
 
+Values _MAY_ be prefixed by a single [Type Annotation](#type-annotation).
+
+### Type Annotation
+
+A type annotation is a prefix to any [Value](#value) that includes a
+_suggestion_ of what type the value is _intended_ to be treated as.
+
+Type annotations are written as a set of `(` and `)` with a single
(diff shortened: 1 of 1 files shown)


tabatkins:
I voted for `tag>value` in the Twitter poll but immediately changed my mind - this `(tag)value` syntax is way better.

Reviewing the commit, +1 from me. The only thing I'm not happiest about is the reserved `uri` type; the distinction between URI and URL is academic and worthless in practice, and "URL" has decidedly won in common parlance.

passcod:
Are non-bare identifiers valid as types?

```
node val=("birth date")1987
```

tabatkins:
Yes:

```
identifier := string | bare-identifier
[...]
type := '(' identifier ')'
```

passcod:
Sorry, wording: I meant, should they be?

tabatkins:
Ah, kk. Yeah, why shouldn't they be? Allowing strings doesn't affect the readability, and the feature itself doesn't seem to have any reason to tie itself to the bare-ident restrictions.

zkat:
string-based identifiers are acceptable intentionally. As Tab says, there doesn't seem to be any reason to limit it to bare identifiers.

zkat:
@tabatkins the reason for calling it "uri" is because that's what JSON Schema calls it, and thus what KDL Schema calls it. I'd be more than happy to have _both_ `url` and `uri`, though, because I agree with you, but we should do that as a separate PR because the Schema spec needs to be updated too.

PR #129, opened 2021-09-03
fix typos

joliss:


zkat (approved):
oh hey nice catch(es)! Thanks for the PR! ^_^

Issue #128, opened 2021-09-03
Minor nitpick: commented-block grammar


larsgw:
Current grammar for `commented-block`:

```
commented-block := ('*' [^\/] | [^*])*
```

As a regex, this would match `**/` even though it should not. In a regex the solution would be a lookahead I think (also for the case `*`, but I am not sure how best to represent that in this grammar syntax. `('*'+ [^*\/] | [^*])*` *would* work for the first case but not for `**` and `*` (or any trailing `*` for that matter).

zkat:
ughghhghhhhhh nice catch. This is a frustrating one. We'll probably have to rewrite multi-line-comment to fix this. Do you want to give this a shot, or should I?

larsgw:
I think `multi-line-comment` is wrong too, the nested `multiline-comment` should be able to go in between text from `commented-block`:

```
multi-line-comment := '/*' (commented-block | multi-line-comment) '*/'
```

If lookaheads are allowed it would become something like this:

```
multi-line-comment := '/*' (commented-block | multi-line-comment)* '*/'
commented-block := ('*' (?=[^\/]) | [^*])+
```

zkat:
I'm trying to remember if any other terms actually require lookahead

larsgw:
For kdljs I think I am using something like this

```
multi-line-comment := '/*' (multi-line-comment | commented-block)*? '*/'
commented-block := '*' | '/' | [^*/]+
```

With the question mark in `multi-line-comment` being a lazy modifier, and the order changed to prefer `multi-line-comment` over `commented-block` in each "iteration".

larsgw:
I limited myself to regex too much. Does this work?

```
multi-line-comment := '/*' commented-block
commented-block := '*/' | (multi-line-comment | '*' | '/' | [^*/]+) commented-block
```

larsgw:
The PEG.js equivalent seems to work so I'll make a PR

```
multi_line_comment = '/*' commented_block
commented_block = '*/' / (multi_line_comment / '*' / '/' / [^*/]+) commented_block
```

PR #130, opened 2021-09-05
Fix multi-line-comment grammar

larsgw:
Fix https://github.com/kdl-org/kdl/issues/128


zkat (approved):

zkat:
You'll want to fix the conflict but this otherwise looks good :)

PR #134, opened 2021-09-09
Add some more reserved type annotations

ericsampson:
I think it would be useful to get these few more common data types into 
For the decimal floating types, I referenced [this protobuf discussion](https://github.com/protocolbuffers/protobuf/pull/7039#issuecomment-679083593) and [this](https://github.com/googleapis/googleapis/blob/master/google/type/decimal.proto)

I'm not sure what to call out for the currency format, I can't find a standard for that yet. There's a [protobuf money type](https://github.com/googleapis/googleapis/blob/master/google/type/money.proto) that's basically a ISO 4217 currency code plus a decimal number, but I don't really want to invent a suggested money format if there's a real one out there somewhere.
For this PR, can we just leave it TBD in order to reserve the type keyword?

### DIFF
--- a/SCHEMA-SPEC.md
+++ b/SCHEMA-SPEC.md
@@ -252,6 +252,12 @@ and property names when the `node-names` or `prop-names` options are activated.
     * `date-time`: ISO8601 date/time format.
     * `time`: "Time" section of ISO8601.
     * `date`: "Date" section of ISO8601.
+    * `duration`: ISO8601 duration format. 
+    * `decimal`: IEEE 754-2008 decimal string format.
+    * `currency`: ISO 4217 currency code.
+    * `country-2`: ISO 3166-1 alpha-2 country code.
+    * `country-3`: ISO 3166-1 alpha-3 country code.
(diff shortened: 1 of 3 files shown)


zkat:
I'd rather just not have the money type if there's no standard we can "hook" into.

Also, if you're adding these types, can you make sure to also update the Schema?

The other types seem fine to me.

ericsampson:
@zkat, am I seeing things right that the schema-spec.md and the schema itself are both missing format spec information for all the numeric types?

zkat:
I guess so. Do you want to port over the formats as a `format` node for numeric values?

ericsampson:
I can do that.

ericsampson:
@zkat I think this is all done and ready for review now, thanks!

zkat:
Can you please resolve the merge conflict?

ericsampson:
Thanks kat, sorry that I didn’t get a chance to fix this, got busy with personal family stuff.

zkat:
it's totally cool. Stuff comes up and it was minor enough to fix myself when I actually took a look :)

PR #136, opened 2021-09-09
use KDL Query for refs

zkat:
We already have a query language. Let's actually use it!

### DIFF
--- a/examples/kdl-schema.kdl
+++ b/examples/kdl-schema.kdl
@@ -20,7 +20,7 @@ document {
         max 1
         children id="node-children" {
             node "node-names" id="node-names-node" description="Validations to apply specifically to arbitrary node names" {
-                children ref="#validations"
+                children ref=r"[id="validations"]"
             }
             node "other-nodes-allowed" id="other-nodes-allowed-node" description="Whether to allow child nodes other than the ones explicitly listed. Defaults to 'false'." {
                 max 1
@@ -31,7 +31,7 @@ document {
                 }
             }
             node "tag-names" description="Validations to apply specifically to arbitrary type tag names" {
-                children ref="#validations"
+                children ref=r"[id="validations"]"
             }
             node "other-tags-allowed" description="Whether to allow child node tags other than the ones explicitly listed. Defaults to 'false'." {
                 max 1
@@ -59,7 +59,7 @@ document {
                             min 1
                             max 1
                         }
-                        prop ref="#info-lang"
+                        prop ref=r"[id="info-lang"]"
                     }
                     node "author" description="Author of the schema" {
                         value id="info-person-name" description="Person name" {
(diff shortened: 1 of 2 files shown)


examples/kdl-schema.kdl
```diff
@@ -20,7 +20,7 @@ document {
         max 1
         children id="node-children" {
             node "node-names" id="node-names-node" description="Validations to apply specifically to arbitrary node names" {
-                children ref="#validations"
+                children ref=r"[id="validations"]"
```
larsgw, line 23: I think this should be `r#"[id="validations"]"#` (same with the other lines).

zkat: omg you're right. facepalm. And I thought I was being so clever lol

larsgw:
I was going to say that it LGTM otherwise, but `ref` should only be able to refer to a single node, no? I guess it's the schema author's problem if it does not, so maybe that's fine. I also wonder how this interacts with linked data and using URL anchors but TBF I think that is just as well-defined as in JSON-LD I think.

zkat:
@larsgw I don't know much about JSON-LD! Let me know if there's any changes you want me to pick up from there.

As far as single-node: I think it's fair to assume that these queries are based on "first result"/"only result". Do you think that needs to be specified?

larsgw:
> I don't know much about JSON-LD! Let me know if there's any changes you want me to pick up from there.

Sorry, I mixed things up again. My chain of thought was that previously `ref` seemed like a relative URL to a node with that `id`, and now it's a different kind of reference. Where I went wrong is that this doesn't matter at all for linked data because (1) it's not like a URL with an id in the fragment would explicitly resolve to a specific KDL node anyway and (2) `ref` is only used for within the schema, not to outside it.

About linked data, I am thinking of using schemas for [RDF](https://en.wikipedia.org/wiki/Resource_Description_Framework)-like linked data, which requires URIs for defining types of resources and properties to link between resources. One way I have seen that be done is by picking a persistent URL for a certain version of the schema and using the URL fragment to point to a specific class or property. So for [CSL-JSON](https://citationstyles.org) the schema URL is `https://resource.citationstyles.org/schema/latest/input/json/csl-data.json` and I would point to the `author` property with `https://resource.citationstyles.org/schema/latest/input/json/csl-data.json#author`. If the schema URL was an HTML page it would scroll directly to the element with `id="author"` so you could say it actually points to that HTML node, but in JSON that doesn't feel as defined.

So I actually meant the concept of using JSON schemas for linked data, not JSON-LD. That is a JSON way of describing RDF (maybe a superset?) and I think I've also seen cases where URL fragments are used to point to entities within a JSON-LD document. The idea is that e.g. `https://example.org/concept/kdl` can return a JSON-LD document with multiple entities (one for KDL the language, one for KDL the project, and one for KDL the specification or something) and each of those entities can get unique URIs (`https://example.org/concept/kdl#language`, `https://example.org/concept/kdl#project`, etc.) without having to redirect those to `https://example.org/concept/kdl` by yourself. Which is also very useful if you are hosting a static site or cannot easily add such redirects for another reason.

> As far as single-node: I think it's fair to assume that these queries are based on "first result"/"only result". Do you think that needs to be specified?

Probably not, that sounds good.

PR #153, opened 2021-09-14
Correct bom from U+FFEF to U+FEFF

chris-morgan:


zkat:
oh no. Oops. Let's just call this a spec bug and pretend it never happened 😂

chris-morgan:
I question the choice to support BOM as a whole, even if it’s only to ignore it—I’d consider it largely a historical mistake of no relevance now—but at least now it’s the right character! 🙂

chris-morgan:
Implementations should also be checked, I imagine.

PR #156, opened 2021-09-14
Added Elixir implementation to list

IceDragon200:
As requested, added my elixir implementation to the list, I'll go add the website one shortly


zkat:
I love the name so much btw

PR #161, opened 2021-09-15
Fix typo

jihchi:
I found a few typo in the query spec document.


zkat (approved):
Nice catch. Thank you!

PR #162, opened 2021-09-15
Prevent markdown from parsing []() as link

benjreinhart:
Markdown was evaluating `[]()` as a link and not showing those characters under the "Non-identifier characters" section. Wrapping it in a code block fixes this and IMO makes the set of characters more obvious (plus, consistent with values above it).


zkat:
oh hah. Nice catch. Thanks!

PR #181, opened 2021-09-19
Add type annotation tests from kdl4j

hkolbeck:
This also tweaks an existing roundtrip test which was modified in https://github.com/kdl-org/kdl/pull/178, `unusual_bare_id_chars_in_quoted_id`. My original intent was essentially to verify that "unusual" chars that were legal in a bare id were correctly printed into a bare id, this shifts back to that intent.


zkat:
Fuck yes, tests!

As an aside, I hope that you working on this stuff yourself means you're feeling a bit better. 💚

PR #184, opened 2021-09-19
Update line continuation wording

benjreinhart:
This commit updates the spec's wording around line continuations, which currently says that the `\` character must have one or more whitespace characters preceding it. That seems to conflict with the grammar definition (i.e., `node-space := ws* escline ws* | ws+`) and behavior of the kdl-rs library.


hkolbeck:
This change matches my understanding of the spec.

zkat (approved):

zkat:
I generally feel like the grammar takes precedent over the prose, so

PR #189, opened 2021-09-20
tests: add more coverage for keywords

benjreinhart:
Hey all,

I noticed that bare identifiers with a prefix that matches a keyword seem to confuse both the JS and Rust implementations into failing to parse. I don't think that's the desired behavior, so I added the following test cases that would be failing for those implementations.

Happy to change the tests around if folks have opinions or remove/merge some if you think any are redundant.

For example, the following two lines each fail:

```kdl
null_id
node null_id=1
```

But these pass:

```kdl
nnull_id
node nnull_id=1
```


zkat (approved):
oh snap. Nice catch. Please be sure to file individual issues for those parsers. this should be fixed for sure.

PR #192, opened 2021-09-21
Fix and simplify identifier character grammar

### LINKED ISSUE #191: The grammar does not exclude some codepoints below 0x20 within a bare identifier
The spec reads:

> The following characters cannot be used anywhere in a bare Identifier:
> - Any codepoint with hexadecimal value `0x20` or below.
> - (snip)

while `idenftifier-char` is defined as follows:

```
identifier-char := unicode - linespace - [\/(){}<>;[]=,"]
linespace := newline | ws | single-line-comment
newline := See Table (All line-break white_space)
ws := bom | unicode-space | multi-line-comment
unicode-space := See Table (All White_Space unicode characters which are not `newline`)
```

`unicode` isn't defined in the grammar but presumably `[\x00-\u{10FFFF}]`. `[\x00-\x08\x0E-\x1F]` are not White_Space nor in the table, so `identifier-char` matches them.

larsgw:
- remove some of the "minus" syntax: it is only
    used in this part of the grammar AFAIK, and
    from my own experience, it can be confusing
  - add \x00-\x20 to the list of non-identifier chars

Fix #191

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -427,8 +427,9 @@ node-space := ws* escline ws* | ws+
 node-terminator := single-line-comment | newline | ';' | eof
 
 identifier := string | bare-identifier
-bare-identifier := ((identifier-char - digit - sign) identifier-char* | sign ((identifier-char - digit) identifier-char*)?) - keyword
-identifier-char := unicode - linespace - [\/(){}<>;[]=,"]
+bare-identifier := (identifier-initial identifier-char* | sign ((identifier-initial | sign) identifier-char*)?) - keyword
+identifier-char := identifier-initial | digit | sign
+identifier-initial := [A-Za-z] | [\x80-\u{10FFFF}] | [!#$%&'*.:?@^_`|~]
 keyword := boolean | 'null'
 prop := identifier '=' value
 value := type? (string | number | keyword)


larsgw:
**Important: this also removes `0x7F`, but I can put that back**

CAD97:
Related note: due to the requirement for the document to be UTF-8, `unicode` potentially is meant to refer to [Unicode Scalar Values](https://unicode.org/glossary/#unicode_scalar_value); that is, `[\0-\u{D7FF}\u{E000}-\u{10FFFF}]`, not `[\0-\u{10FFFF}]` (see #207).

zkat:
Moving this to target the v2 branch since it includes breaking changes.

tabatkins:
> Related note: due to the requirement for the document to be UTF-8, unicode potentially is meant to refer to Unicode Scalar Values; that is, [\0-\u{D7FF}\u{E000}-\u{10FFFF}], not [\0-\u{10FFFF}] (see #207).

Since the requirement for UTF-8 already exists, this distinction is moot; you can't validly encode the surrogate codepoints into UTF-8 anyway.

tabatkins:
This change (almost certainly unintentionally) would allow non-ASCII linespace as valid ident chars, which I think would be a bad idea.

PR #193, opened 2021-09-21
fix: incorrect expected test output

danini-the-panini:


hkolbeck:
The temp file is definitely my bad (and the one kdl4j uses in its test suite since it does print using the same radix), but the existing file with the decimal representation is correct for this repo given this from the test README:

```
- All numbers must be converted to their simplest decimal representation. That means that hex, octal, and binary must all be converted to decimals. All floats must be represented using E notation, with a single digit left of the decimal point if the float is less than 1. While parsers are required to consume different number syntaxes, they are under no obligation to represent numbers in any particular way.
```

Implementations MAY track radix and roundtrip it, but they aren't forced to, so the standard test suite doesn't expect it.

danini-the-panini:
I've made #196 proposing a more appropriate fix in accordance to the documentation.

PR #194, opened 2021-09-21
fix: unescaped / in quoted type tests

danini-the-panini:


zkat:
Interesting. You shouldn't have to escape / in regular strings?

hkolbeck:
I think `/` is weird in that it can be escaped, but doesn't have to be in other places as well. I think it's fine unescaped here.

danini-the-panini:
I've proposed an alternate solution here: #197

danini-the-panini:
Alternatively, #198

PR #196, opened 2021-09-22
fix: non-decimal test output according to docs

danini-the-panini:
According to the tests README:

> All numbers must be converted to their simplest decimal representation. That means that hex, octal, and binary must all be converted to decimals.

Closes #193


zkat (approved):

zkat:
Thanks for all these patches! Sorry it took me a while to get around to them :)

PR #197, opened 2021-09-22
Do not escape / (Solidus, Forwardslash)

danini-the-panini:
I don't believe `/` need to be escaped in strings. This may be considered a breaking change now that we've reached 1.0, so another solution might be to allow it to be escaped during parsing, but not to escape it when stringifying.

Closes #194


zkat:
This is, unfortunately, a breaking change, so I'm moving it to target the v2 branch.

tabatkins:
The secondary solution (allow it to be escaped in parsing, but don't require it to be escaped in printing) appears to already be expected by the testsuite; see [`all_escapes.kdl`](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/input/all_escapes.kdl) and its [expected output](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/expected_kdl/all_escapes.kdl).

(I think superfluous escapes are fine - all it means is we can't give special meaning to an escaped forward slash `\/` later -  so I don't think we *need* to change any behavior here at all. But it would certainly be *tidier* to remove `\/`, since it *does* suggest that `/` would do something special in strings if not escaped.)

PR #206, opened 2021-09-27
argument/property ordering clarification

### LINKED ISSUE #205: Does argument order w.r.t. properties matter?
The spec says that

> Arguments are ordered relative to each other and that order must be preserved in order to maintain the semantics.
>
> By contrast, Property order _SHOULD NOT_ matter to implementations. [Children](https://github.com/kdl-org/kdl/blob/main/SPEC.md#children-block) should be used if an order-sensitive key/value data structure must be represented in KDL.

My question is: if a property is between two arguments, is that allowed to impact the semantics of the arguments in a fully-SHOULD-compliant implementation?

Using the given example,

```kdl
foo 1 key="val" 3 {
    bar
    (role)baz 1 2
}
// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// is that semantically different from
// vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
foo 1 3 key="val" {
    bar
    (role)baz 1 2
}
```

More informally, does the spec require grouping the above node as roughly `foo (1) (key="val") (3)`, or is an implementation allowed to interpret it as `foo (1) key=("val" 3)`?

If argument/property mixing SHOULD NOT matter to implementations, I suggest the following clarification to the spec:

```diff
@@ -60,2 +60,2 @@
-Arguments are ordered relative to each other and that order must be preserved
-in order to maintain the semantics.
+Arguments are ordered relative to each other (but not Properties), and that
+order must be preserved in order to maintain the semantics.
```

If argument/property mixing MAY matter to implementations, I suggest the following clarification to the spec:

```diff
@@ -60,6 +60,6 @@
-Arguments are ordered relative to each other and that order must be preserved
-in order to maintain the semantics.
+Arguments are ordered relative to each other and to Properties, and that
+order must be preserved in order to maintain the semantics.

-By contrast, Property order _SHOULD NOT_ matter to implementations.
-[Children](#children-block) should be used if an order-sensitive key/value
-data structure must be represented in KDL.
+By contrast, Property order relative to each other _SHOULD NOT_ matter
+to implementations. [Children](#children-block) should be used if an
+order-sensitive key/value data structure must be represented in KDL.
```

-----

The scope of this question is that I'm drafting out a semi-formal spec for Serde-in-KDL (SiK) a la JiK or XiK, describing how to map between the [Serde data model](https://serde.rs/data-model.html) and KDL. [Current draft here](https://github.com/CAD97/serde-kdl/). I want SiK to feel like natural KDL use as much as possible, and I think I've captured that, save for one current limitation: mixing properties and arguments in a single node isn't supported. In the document I have a potential way to support arguments before properties, but whether this is possible depends on whether this is deserializable within serde's API (unknown, slightly positive), and arguments mixed with properties will need to know the answer to the above spec clarification before I can even consider deciding what the semantics of such should be.

And yes, I am [reaching out to collaborate with Lucretiel/kaydle](https://github.com/Lucretiel/kaydle/issues/9), now that I'm aware of that effort.

CAD97:
Closes #205.

This does not change the meaning of the spec; it just makes one interaction a bit more explicit.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -57,8 +57,8 @@ slash-escaped line continuation](#line-continuation). Arguments and Properties
 may be interspersed in any order, much like is common with positional
 arguments vs options in command line tools.
 
-Arguments are ordered relative to each other and that order must be preserved
-in order to maintain the semantics.
+Arguments are ordered relative to each other (but not relative to Properties)
+and that order must be preserved in order to maintain the semantics.
 
 By contrast, Property order _SHOULD NOT_ matter to implementations.
 [Children](#children-block) should be used if an order-sensitive key/value


zkat (approved):
I mean I thought this was _pretty clear_, but I think doubling-down here wouldn't hurt. I'll take it. :)

PR #219, opened 2021-10-05
Empty children aren't reflected in the data model.

### LINKED ISSUE #218: Differentiation between "no children" and "empty children"
I'm wondering if implementations should be explicit about the difference between the following two documents:

```
node
```

```
node {
}
```

At the moment, I'm interpreting them identically. When it comes time to serialise the document, both inputs would be encoded as the first example. Is this correct?

tabatkins:
Closes #218


hkolbeck:
Can you add a note to the test README as well?

tabatkins:
That's already in the README due to

> * Node fields should be `identifier <values> <properties> <children only if non-empty>`

right? That's what made me think that an empty child list should be omitted in the first place.

hkolbeck:
Whoops, yep. LGTM

PR #221, opened 2021-10-06
KQL: require operator and change operator grammar a bit

zkat:
Ref: https://github.com/kdl-org/kdl/pull/204#issuecomment-935378397

This just makes it so "operators" are required, meaning we don't have to worry about syntax ambiguity with bare identifiers anymore :)


zkat:
cc @tabatkins @larsgw I'd love your feedback

larsgw:
Slightly related, did you see this page from the CSS WG being tweeted yesterday? https://wiki.csswg.org/ideas/mistakes

> # Incomplete List of Mistakes in the Design of CSS
> ...
> * descendant combinator should have been » and indirect sibling combinator should have been ++, so there's some logical relationships among the selectors' ascii art

PR #222, opened 2021-10-06
KQL: remove map operator and accessors

zkat:
Honestly, they're just too implementation-specific

### DIFF
--- a/QUERY-SPEC.md
+++ b/QUERY-SPEC.md
@@ -69,33 +69,6 @@ is not one of those, the matcher will always fail:
 
 * `[val() = (foo)]`: Selects any element whose tag is "foo".
 
-## Map Operator
-
-KQL implementations MAY support a "map operator", `=>`, that allows selection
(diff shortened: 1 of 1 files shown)


danini-the-panini:
My poor heart when I read this just as I finally finished getting my map operator and accessor implementations working 😭

zkat:
Map operator is still great, and if implementations want something like that, it makes sense!

I just didn't think it made sense for the _spec_

fabiancook:
The individual accessors without the map operator line up very nicely to JSX land concepts. 

The map operator on the back of the accessors works really well too. 

Since they have stayed in the spec this long, and map operators are optional now, instead of removing them both, maybe we should just map accessors optional too? 

I have included both accessors and map operator support in the JSX query engine. 

https://github.com/virtualstate/kdl/blob/464d0cd5e2aae99bdcc97f00ea8a50ce00a4021a/src/tests/readme.tsx#L79-L88

Issue #225, opened 2021-10-06
KDL specification does not explicitly define a data model but only a syntax that implies some data model of KDL. As long as you parse from KDL syntax into some data structure of a programming language, a KDL data model is not needed. In contrast both KQL and KDL Schema build on a data model. The lack of its explicit definition leaves open handling of some edge cases, such as:


zkat:
KDL specification does not explicitly define a data model but only a syntax that implies some data model of KDL. As long as you parse from KDL syntax into some data structure of a programming language, a KDL data model is not needed. In contrast both KQL and KDL Schema build on a data model. The lack of its explicit definition leaves open handling of some edge cases, such as:

* Is an empty string tag different to no tag ( does `("")value` differ from `value`)?
* Is an empty list of child nodes different from no child nodes (does `node {}` differ from `node`)?

The data model should abstract from syntax details such as comments, how nodes are terminated, whether names are given as bare identifiers or given as strings etc.

Here is a draft of a possible data model (without explicit definition of *name* and *value* and both questions above answered with no):

* a *document* is a list of *nodes*
* a *node* consist of
  * a *tag* being a *name* 
  * a *name*
  * a list of *arguments*, each being a *value*
  * a set of *properties*, each consisting of a *name* and *value*, and each *name* being unique
  * a list of *children* being a *document*

_Originally posted by @nichtich in https://github.com/kdl-org/kdl/discussions/183_

zkat:
@nichtich I like the data model so far. Do you want to start a PR so we can start iterating on it, against the v2 branch?

nichtich:
The KDL Specification briefly refers to "the [intended] data model" in the introduction. This is *another* data model, so I would remove these words from the KDL Specification to avoid confusion.

As hopefully made clear in the Application Notes, KDL data model should not be mandatory part of KDL Specification. Implementation will likely restrict at least number representation anyway.

PR #228, opened 2021-10-07
Start draft of KDL Data Model

nichtich:
Thanks for discussion (<https://github.com/kdl-org/kdl/issues/225>). In addition to further changes to the specification of KDL Data model, the document should be referenced from other documents in this repository.


zkat:
Shall we roll this up as part of KDL 2.0?

PR #233, opened 2021-10-08
Empty tag is same as no tag.

tabatkins:
Don't merge until data model updates (#228) are approved and merged.


djmattyg007:
Is this to be considered part of the 2.0 spec? My understanding of the 1.0 spec is that there is absolutely a difference between an empty tag and no tag.

zkat:
I'm not really into this. I don't want us to start delving into falsiness? I probably missed the conversation about this or forgot about it tho

PR #234, opened 2021-10-11
Allow "empty" single line comments in the spec

basile-henry:
As I read the grammar in the spec, `"//"` wouldn't parse as a single-line-comment as it requires as least one non-newline character after the slashes.


CAD97:
If it wouldn't parse successfully otherwise, it's not breaking, as it's solely extending the allowed grammar. I don't believe there's any case where `//` is valid currently, where it would potentially be a comment after this.

zkat (approved):

PR #236, opened 2021-10-13
Define `digit` in formal grammar

CAD97:
`digit` is used as a subtraction in bare-identifier, but never defined.

--- 

An alternate change:

```
decimal := integer ('.' digit digit_*)? exponent?
exponent := ('e' | 'E') integer
integer := sign? digit digit_*
digit := [0-9]
digit_ := [0-9_]
sign := '+' | '-'
```

(NOTE: this has now flip-flopped to opposite what it was when @tabatkins made their comment.)

I think I slightly prefer the PR'd version, but it changes the implied grammar tree structure more significantly. This change "just" gives names to the `[0-9]` and `[0-9_]` productions.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -446,9 +446,10 @@ raw-string-quotes := '"' .* '"'
 
 number := decimal | hex | octal | binary
 
-decimal := integer ('.' [0-9] [0-9_]*)? exponent?
-exponent := ('e' | 'E') integer
-integer := sign? [0-9] [0-9_]*
+decimal := sign? integer ('.' integer)? exponent?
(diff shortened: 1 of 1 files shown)


tabatkins:
Afaict both the PR and the alternate change are correct and produce no normative change beyond solving the original issue 👍 

I actually slightly prefer the alternate change; I know that I initially missed that an exponent could have a sign when I was implementing, and so I like being more explicit with where exactly a sign is allowed in a number. It also avoids defining a `digit_`, which I slightly prefer.

CAD97:
Swapped in the preferred change instead of the less desirable minimal one

PR #239, opened 2021-10-16
Add link to additional python implementation

djmattyg007:
I created this implementation by forking the now-defunct original kdl-py implementation. It differs from the existing implementation in a few important ways:

- It provides a more python-native interface akin to the existing JSON and YAML packages
- It provides powerful support for custom types, while ensuring support for built-in types is first-class
- It deliberately doesn't provide as much control over the encoding process as kdl-py does


djmattyg007:
I started working on `cuddle` almost a month ago, shortly after the announcement of KDL 1.0 being formalised. At the time, the only implementation I was aware of was the broken version of `kdl-py`.

tabatkins:
Nice, my kdl-py was an entirely fresh impl just *after* the 1.0 announcement, because I saw the existing py impl was defunct 😀

PR #271, opened 2022-04-25
Add more info about the hyphen character in identifiers

pingiun:
This caused confusion for me when making my implementation as explained in: https://github.com/kdl-org/kdl/discussions/200#discussioncomment-2628423

The exact rules in the full grammar are very precise but this adds some explicit information in the prose section


tabatkins:
Looks reasonable to me!

Issue #51, opened 2020-12-26
.NET implementation


zkat:
We should have a nice .net parser for kdl. Hopefully we can also add it to the org!

jtcooper10:
It looks like [Andrew Hall on Twitter](https://twitter.com/ryzngard/status/1341912796171145216) has started on one, but either way I'd be happy to lend a hand with a C# implementation. [Sprache](https://github.com/sprache/Sprache/tree/develop) seems like a good place to start.

jkoritzinsky:
Ive had some musings about using SDLang or KDL as an alternative markup language for Avalonia, so I might take a stab at an implementation (non-Sprache based for perf reasons) some time.

chan-w:
I'm working on a [FParsec](https://www.quanttec.com/fparsec/) based parser in F#: https://github.com/chan-w/kdl-dotnet

borland:
I've ported kdl4j (which is great, btw) across from java into a [.NET library](https://github.com/borland/kdl-net)

The port is mostly literal but I've tweaked it very slightly to fit C# better. It's not quite complete (I still need to add floating point number support as at 15 jan 2021) but everything else is there, and has good test coverage thanks to kdl4j

Aside: I hadn't seen this issue earlier. I didn't realize Andrew hall was also doing one

zkat:
@borland @chan-w did anything ever come of your implementations? :)

borland:
@zkat not that I'm aware of. 7 people starred it on github, and I had some fun chats with people at work.

We did actually consider using it for configuring real asp.net core apps, rather than JSON. Everyone thought it was cool, and a nice improvement - but ultimately we didn't care enough about config to warrant taking on a highly experimental new language for that.

oledfish:
I have been working on my own implementation, also based around kdl4j though not a direct port of it, available [here](https://github.com/oledfish/Kadlet). It hasn't been published on NuGet yet.

I feel as if it's ready to be used, though I am a bit lost as to how implementations are declared to be good enough or whether I could get some feedback. Either way I hope it can be useful.

borland:
Did you see my one? You're welcome to take code and testcases and stuff
from it.

I don't think it was listed on the main KDL page; as with you I'm not sure
how that happens, or what criteria may apply first.
borland/kdl-net: A C# parser for the KDL document language (github.com)
<https://github.com/borland/kdl-net>

Kind regards

Orion

oledfish:
[...]

PR #277, opened 2022-07-23
Add .NET library to the implementation list

oledfish:
As per #51


zkat (approved):
awesome, thank you!

Do you mind doing this for https://github.com/kdl-org/kdl-org.github.io/blob/main/src/_includes/partials/implementations.md as well? That'll put it up on the site too.

PR #278, opened 2022-07-29
Add @virtualstate/kdl to implementations

fabiancook:
This includes [@virtualstate/kdl](https://github.com/virtualstate/kdl) in the implementation list

Currently the primary focus of the module is to provide a query syntax for JSX nodes, using KDL Query Language

e.g. 

```typescript jsx
import {prepare} from "@virtualstate/kdl";

const node = (
    <main>
        <h1>@virtualstate/focus</h1>
        <blockquote>Version: <span>1.0.0</span></blockquote>
    </main>
);
const [span] = await prepare(
    node,
    `main blockquote > span`
);
console.log(span); // Is the node for <span>1.0.0</span>
```


zkat:
Please remember to also make a PR to the website to add it there!

PR #285, opened 2022-08-21
Update formal grammar for KDL 2.0

### LINKED ISSUE #284: Spec is unclear about how slashdash comments work
Consider

```
node/-"val""val"
node /-{} "val"
node /- /-{} {}
```

kdl-rs accepts this, but the formal grammar forbids all three nodes here.

EDIT: preferred resolution is now #285.

<details><summary>Original suggestion</summary>

My preferred resolution at the moment[^1] is to make the implementation and prose match the formal grammar here, with one small change: require whitespace before a slashdashed children block as well. (The lack of such causes implementation issues[^2].)

[^1]: And what I'm implementing for the time being in my annoyingly-zero-copy parse/visitor. Don't let me strongarm you, though; I've figured out a reasonable (though kind of annoying) way to implement the reference implementation's behavior, and that's the most interesting option to implement in my LL(2) recursive descent parse/visitor. In fact, most of the implementation annoyance is still there anyway in order to support a non-spaced non-slashed children block while requiring spacing for everything else.

[^2]: Namely: the grammar is no longer LL(n), as it requires unbounded *token* lookahead when seeing a slashdash (past arbitrarily many escaped lines) to determine if it's a slashdashed argument/property (error) or a slashdashed children block (allowed). This arbitrary lookahead isn't just an implementation concern, either; it means that the location of the produced error is potentially far from the cause of the error.

Suggested wording:

In Argument/Property, delete the references to `/-`. The Children Block's prose does not currently allow for it to be slashdashed, despite this being allowed in the formal grammar and reference implementation. In Node:

> Nodes *MAY* be prefixed with `/-` to "comment out" the entire node, including its properties, arguments, and children, making it act as whitespace, even if it spreads across multiple lines. Additionally, the node's properties, arguments, and children block *MAY* themselves be prefixed with `/-` to "comment out" that entire component. If the children block is "commented out" in this fashion, it *MUST* be separated by whitespace or a slash-escaped line continuation. Even if a component is "commented out" in this fashion, it still acts as the commented out component *syntactically*. (For example, "commented out" arguments and properties must still be separated by plain whitespace, and if a "commented out" children is present, it must be after all arguments and properties, and there may not be another children block, whether "commented out" or not.)

In the formal grammar:

> ```
> node := ('/-' node-space*)? type? identifier node-prop-or-arg* node-children? node-space* node-terminator
> node-prop-or-arg := node-space+ ('/-' node-space*)? (prop | value)
> node-children := (node-space+ '/-') node-space* '{' nodes '}'
> ```

</details>

CAD97:
Closes #284 by making the formal grammar match the implementation and prose, explicitly treating slashdashed nodes as line-space and slashdashed node properties, arguments, and children as node-space.


zkat (approved):
Honestly I really like how much this simplifies the slashdash implementation, which was kind of tricky before.

zkat:
The formal grammar is actually the "authoritative" version here, rather than the prose, so this is basically a breaking change. I think it's about time we start talking about the next version of KDL, though. We have a few breaking changes in the pipeline that have just been sitting on the sidelines (mostly minor things, tho).

SPEC.md
```diff
@@ -442,17 +442,19 @@ Note that for the purpose of new lines, CRLF is considered _a single newline_.
...
 ```
-nodes := linespace* (node nodes?)? linespace*
+nodes := (line-space* node)* line-space*
 
-node := ('/-' node-space*)? type? identifier (node-space+ node-prop-or-arg)* (node-space* node-children ws*)? node-space* node-terminator
-node-prop-or-arg := ('/-' node-space*)? (prop | value)
-node-children := ('/-' node-space*)? '{' nodes '}'
-node-space := ws* escline ws* | ws+
+line-space := newline | ws | single-line-comment | '/-' node-space* node
+node-space := ws* escline ws* | ws+ | '/-' node-space* (node-prop-or-arg | node-children)
```
CAD97, line 448: Given that we're treating this as breaking anyway, I want to propose writing this instead as

```suggestion
node-space := plain-node-space+ ('/-' plain-node-space* (node-prop-or-arg | node-children) plain-node-space+)
plain-node-space := ws* escline ws* | ws+
```

in order to require separating whitespace around slashdashes (even though that'll make my implementation more annoying, I think).

This split is actually necessary to avoid allowing `/- /-{} {}`.... which kdl-rs also allows 🙃

CAD97:
Also, it's probably worth doing a survey of existing implementations to see what exactly each of their behaviors here is. It's not unlikely there's divergence here.

SPEC.md
CAD97: As written, this does still allow `node /-"val"{}` (which is `node {}`), as space is not required before a children block. I'm still slightly biased towards requiring space before children blocks or never requiring space (for uniformity), but I see the appeal of allowing omitting the space before the children block. An interesting approach, if somewhat involved, is to allow `node{}` but not `node "val"{}`:

```
node := type? identifier node-entries node-space* node-terminator
node-entries := node-children | (node-space+ (prop | value))* (node-space+ node-children)?
```

I kinda like this, but this is drifting further into more involved changes.

zkat: My inclination with all this is to ask "which one is the simplest to implement that will work as expected in most cases?" One of KDL's design goals is to be relatively straightforward to implement, even if it results in some "funny-looking" documents.

CAD97: In that case, I can say that from my experience:

- With the grammar changed to just classify slashdash as whitespace, implementing is fairly simple either way;
- Requiring whitespace only sometimes is more annoying than being consistent (you have to remember if you ate whitespace to decide later if there's an error); and
- Not requiring whitespace is simpler to implement than requiring whitespace, as less code is easier (it's quite literally just delete the code which requires the whitespace). Additionally, never requiring whitespace means that it's possible to use a lexer which discards the trivia (whitespace, comments) eagerly.

The closer the grammar can be to satisfying being a PEG the easier it'll be to implement, as PEG is quite literally just a declarative recursive descent parser.

bgotink: Shouldn't this then also change?

```diff
-line-space := newline | ws | single-line-comment | '/-' node-space* node
+line-space := newline | ws | single-line-comment | '/-' plain-node-space* node
```

Otherwise the following would be valid and imo pretty confusing

```kdl
/- /-"commented" /-{ commented; } commented "commented" { commented; }
```

CAD97: Ah yes, I agree, that's a good change as well, if we're distinguishing plain whitespace.

zkat (commented):
So I guess in summary, there's a few things to change here before we mark it for KDL 2.0:

1. Require spaces before `node-children` in nodes (which I think is perfectly fine to do?
2. Adopt the `plain-node-space` patch in the comment.
3. Make sure we're treating slashdash as whitespace, period.
4. Include the `plain-node-space` change for `line-space` that @bgotink suggested

How's that sound?

zkat:
Additionally, before merging this, I'd like to see a handful of updated tests in our compliance test suite that cover these corner cases.

SPEC.md
```diff
@@ -442,17 +460,22 @@ Note that for the purpose of new lines, CRLF is considered _a single newline_.
...
+plain-node-space := ws* escline ws* | ws+
 
-node := ('/-' node-space*)? type? identifier (node-space+ node-prop-or-arg)* (node-space* node-children ws*)? node-space* node-terminator
-node-prop-or-arg := ('/-' node-space*)? (prop | value)
-node-children := ('/-' node-space*)? '{' nodes '}'
-node-space := ws* escline ws* | ws+
+line-space := (plain-line-space+ '/-' plain-node-space* node)* plain-line-space+
+node-space := (plain-node-space+ '/-' plain-node-space* (node-prop-or-arg | node-children))* plain-node-space+
+
+node := type? identifier (node-space+ node-prop-or-arg)* (node-space+ node-children)? node-space* node-terminator
```
CAD97, line 471: One remaining spec bug we should iron out: currently

```kdl
node /-"argument";
```

isn't valid, as `node-space` requires trailing `plain-node-space`. The best fix I can think of is to take the trailing `plain-node-space` out of `node-space` and put it into `node`, e.g.

```suggestion
line-space := plain-line-space+ ('/-' plain-node-space* node)?
node-space := plain-node-space+ ('/-' plain-node-space* (node-prop-or-arg | node-children))?

node := type? identifier (node-space* plain-node-space+ node-prop-or-arg)* (node-space* plain-node-space+ node-children)? node-space* node-terminator
```

Thoughts? Checks that we're speccing what we want to be?

CAD97:
I'll bring in some conformance tests for these edge cases later this week; I'm writing a bunch as part of writing my implementation anyway.

SPEC.md
CAD97: An alternative formulation, perhaps better:

```suggestion
line-space := plain-line-space+ ('/-' plain-node-space* node)?
node-space := plain-node-space+ ('/-' plain-node-space* (node-prop-or-arg | node-children))?

required-node-space := node-space* plain-node-space+
optional-node-space := node-space*

node := type? identifier (required-node-space node-prop-or-arg)* (required-node-space node-children)? optional-node-space node-terminator
```

SPEC.md
```diff
@@ -24,6 +26,22 @@ the directions if the data stream were only ASCII text. They do not refer
...
+  blocks do not prevent the presence of later arguments, properties, or children
+  blocks on the attached node.
+
+### Constrained
+
+- Previously, whitespace was not required before a children block, i.e. `node{}`
+  was valid. Now, whitespace is required before a children block, the same as
+  before arguments and properties.
+- `/-` comments on nodes must also be separated by plain (non-`/-`) whitespace. 
+
```
zkat, line 44: Can you put these in the new CHANGELOG.md file?

CAD97: Will do later this week along with the conformance tests (I've been doing this in the web UI so far)

tabatkins:
I merged in the "alternate formulation" that uses `required-node-space` and `optional-node-space`, as it's exactly what I was going to suggest while reading the first fix attempt. ^_^

zkat:
I'm going to merge this, but I kinda want to see what it takes to allow `node{}` to be legal. We can do that after this, though.

Issue #255, opened 2022-02-02
formal identifier grammar doesn't match prose


CAD97:
[The prose](https://github.com/kdl-org/kdl/blob/main/SPEC.md#non-identifier-characters) says that

> Non-identifier characters
>
> The following characters cannot be used anywhere in a bare Identifier:
>
> - Any codepoint with hexadecimal value 0x20 or below.
> - Any codepoint with hexadecimal value higher than 0x10FFFF.
> - Any of `\/(){}<>;[]=,"`

The formal grammar says that

```
identifier-char := unicode - linespace - [\/(){}<>;[]=,"]
linespace := newline | ws | single-line-comment

newline := See Table (All line-break white_space)
unicode-space := See Table (All White_Space unicode characters which are not `newline`)
```

This means that the formal grammar excludes 0x09, 0x0A, 0x0C, 0x0D, and 0x20, but allows the other codepoints < 0x20.

The prose also doesn't specify that all whitespace is non-identifier characters, though this is clearly implied, and doesn't matter as much as it does in the grammar formalization.

zkat:
Happy to take a patch for this!

marrus-sh:
(this is a duplicate of #191)

zkat:
This should be fixed in the kdl-v2 branch

Issue #250, opened 2021-11-04
v2.0 additional restricted literal characters


tabatkins:
Currently, idents [disallow](https://github.com/kdl-org/kdl/blob/main/SPEC.md#non-identifier-characters) a few characters from being expressed literally, requiring they be escaped if authors want to include them:

* codepoints < 0x20 (control characters)
* codepoints > 0x10FFFF (invalid codepoints)
* some ASCII characters reserved for syntax reasons

I think there's a few more we can reasonably restrict to make KDL documents more readable/understandable:

* 0x7F, another ASCII control character
* 0x2066-2069, 0x202A-202E, the [unicode "direction control" characters](https://www.w3.org/International/questions/qa-bidi-unicode-controls)

Removing 0x7F just seems like fixing an omission; it's easy to forget that the ASCII control characters aren't contiguous.

Removing the direction-control characters helps keep KDL source readable; the direction *override* characters in particular are somewhat fraught to show up in plain-text documents, as they can corrupt the display of following text in the wrong direction (as demonstrated in the recent somewhat-hyperbolic complaints about them showing up in Rust and other source languages as a possible review-attack). If these character are desired for use in text values, such as strings, they can still be escaped; their literal usage in what is otherwise an ASCII-based language is virtually always either accidental or malicious, since they're intended for *text formatting* and have no semantic meaning.

The BOM is allowed at the start of a KDL document

(A previous issue suggested restricting the surrogate-pair characters as well (0xD800-DFFF); these are already restricted implicitly by the requirement that KDL documents be encoded in UTF-8, where such codepoints can't be validly encoded. As such I'm continuing to omit them from these suggestions.)

While there are still a number of "invisible" characters in Unicode that could potentially be confusing or accidental, they also have semantic uses, so I don't currently recommend restricting them.

marrus-sh:
I think you should not remove direction characters as they make it impossible to literally encode bidirectional strings, which is important for internationalization. It is true that BIDI control characters can create review‐attacks, but KDL is not a programming language and the probability of someone it needing to encode lengthy strings (which may include bidirectional text) is pretty high. Linters and formatters can be used by individual projects to detect and warn about the use of these characters if needed; there is no reason to forbid them at a language level.

I would suggest disallowing exactly the same characters as [RestrictedChar in XML 1.1](https://www.w3.org/TR/xml11/#NT-RestrictedChar), plus U+0000, U+FFFE, and U+FFFF (which are not allowed to be escaped in XML either).

[...]

Issue #213, opened 2021-09-28
`\` at EOL should escape all whitespace after it and treat it as a single one


zkat:
in one of their streams, @lucretiel pointed out that Rust strings let you do this:

```rust
let foo = "x\
    y\
    z\
    ";
```

I think this is a GREAT compromise for not automatically collecting indentation? I think we should do this too.

My one question is that I'm not 100% sure what Rust itself does, but I think whatever that is, we should probably just copy it.

CAD97:
Rust escapes _all_ whitespace until the next non-whitespace character.

[Specifically,](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b6e7d3fe1671d9ef230e1f97f1cee7bf)

```rust
fn main() {
    dbg!("x\
    y\
    z\
    ");
}
```

```
[src/main.rs:2] "x\
    y\
    z\
    " = "xyz"
```

In strings, this makes sense, but outside of strings, specifically for the purpose of KDL, I think treating `\\\n\p{White_Space}*`<sub>regex</sub> as one unit of whitespace (not linespace) is better from a practicality of use standpoint (no splitting lexer atoms over multiple lines).

Lucretiel:
Just to be clear, this proposal only applies to quoted string literals in KDL; it doesn't change the behavior of linespace / splitting nodes over multiple lines or anything about how KDL handles syntactic whitespace.

The [Rust behavior](https://doc.rust-lang.org/reference/tokens.html#string-literals) is specifically that a backslash preceding a line break (newline or CRLF) consumes itself, the line break, and all succeeding literal whitespace at the beginning of the next line. I'd propose the simpler, more general rule, which is simply that a backslash consume all literal whitespace that succeeds it. In either cases, they notably only consume literal whitespace; they do *not* consume escaped whitespace. This means we can write things like:

```
node "\
    Header\n\
    \tIndented\n\
    \tIndented again\
"
```

This is exactly equivalent to:

```
node "Header
\tIndented
\tIndented again"
```

eugenesvk:
Would a better rule be to consume only the same amount of whitespace, so to cut out the prefix whitespace in vertically aligned columns, but also allow non-prefix whitespace at line beginning
```
node "\
  Header\n\
    - item 1 with 2-spaces indent\n\
    - item 2 with 2-spaces indent\n\
  Continue\
"
```

```
node "Header
  - item 1 with 2-spaces indent
  - item 2 with 2-spaces indent
[...]

PR #286, opened 2022-08-28
Merge KDL v2

zkat:
Here it is! The long-awaited KDL v2, which is where we go ahead and make a handful of technically-breaking changes to address some corner cases we've run into over the past year while KDL has been getting implemented in a bunch of languages by various people.

I'd love to get feedback on what we have slated, and whether there's anything else we should definitely include when this goes out.


CAD97:
I have a *slight* preference for #241 over #204 personally, though only slight.

tests/test_cases/input/no_solidus_escape.kdl
```diff
@@ -0,0 +1 @@
+node "\\"
```
IceDragon200, line 1: Shouldn't there also be an output case for this, and one for backslash then?

Seeing as the escaped form was removed

hkolbeck:
I have a preference for #204, because the primary use case I can see for `#` in bare identifiers is hashtag-like which would be illegal under either, and it seems better to go with the simpler rule.

That preference is not terribly strong, though.

CAD97:
> the primary use case I can see for `#` in bare identifiers is hashtag-like

To clarify, #241 allows `#ident` as a bare ident, and both will of course still allow `"r#ident"` as a quoted ident.

Argument for allowing: transliterating CSS selectors, for e.g. CSS-in-KDL. Argument against allowing: using the syntax in KQL as a selector like CSS.

lilyball:
> Argument for allowing: transliterating CSS selectors, for e.g. CSS-in-KDL. Argument against allowing: using the syntax in KQL as a selector like CSS.

`#foo` in CSS is special-casing the `id` attribute. KQL doesn't have an equivalent to HTML's `id`, and using `#foo` syntax in KQL to mean something else might be confusing given its meaning in CSS, so I don't find the argument against compelling.

My inclination is to prefer #241 as well, as I think being able to write hashtags is neat. It also allows for doing things like writing Nix flake references as bare words, e.g. `nixpkgs#hello`.

Lucretiel:
Can we squeeze https://github.com/kdl-org/kdl/issues/213 into this? The specific proposal is the addition of escaped whitespace in string literals– that `\`, followed by literal (non-escaped) whitespace, should consume and discard all that whitespace. This is a slight simplification of the Rust rule, which specifically requires that `\` be followed by `\n`.

hkolbeck:
I'm also a fan of #213, though it seems like there's some ambiguity in the discussion. Namely, does 
```
- "x\
    y\
    z"
```
Translate to `"xyz"`, `"x y z"`, or `"x\ny\nz"`?

zkat:
> Can we squeeze #213 into this? The specific proposal is the addition of escaped whitespace in string literals– that `\`, followed by literal (non-escaped) whitespace, should consume and discard all that whitespace. This is a slight simplification of the Rust rule, which specifically requires that `\` be followed by `\n`.

@Lucretiel do you have time to put together a PR with this grammar+prose change? I'm game.

Lucretiel:
> > Can we squeeze #213 into this? The specific proposal is the addition of escaped whitespace in string literals– that `\`, followed by literal (non-escaped) whitespace, should consume and discard all that whitespace. This is a slight simplification of the Rust rule, which specifically requires that `\` be followed by `\n`.
> 
> @Lucretiel do you have time to put together a PR with this grammar+prose change? I'm game.

Yes, tonight I can put that together :) should it be in the form of an amendment to SPEC.md?

Lucretiel:
> I'm also a fan of #213, though it seems like there's some ambiguity in the discussion. Namely, does
> 
> ```
> - "x\
>     y\
>     z"
> ```
> 
> Translate to `"xyz"`, `"x y z"`, or `"x\ny\nz"`?

I agree there's some ambiguity in the original. That example would translate to `"xyz"`, because *all* literal whitespace after the `\` is consumed and discarded. If you want to retain whitespace, it should either come before the `\` or itself be escaped. I think my comment (https://github.com/kdl-org/kdl/issues/213#issuecomment-929869117) succinctly describes the

zkat:
> > > Can we squeeze #213 into this? The specific proposal is the addition of escaped whitespace in string literals– that `\`, followed by literal (non-escaped) whitespace, should consume and discard all that whitespace. This is a slight simplification of the Rust rule, which specifically requires that `\` be followed by `\n`.
> > 
> > 
> > @Lucretiel do you have time to put together a PR with this grammar+prose change? I'm game.
> 
> Yes, tonight I can put that together :) should it be in the form of an amendment to SPEC.md?

yep!

> I agree there's some ambiguity in the original. That example would translate to "xyz", because all literal whitespace after the \ is consumed and discarded. If you want to retain whitespace, it should either come before the \ or itself be escaped. I think my comment (https://github.com/kdl-org/kdl/issues/213#issuecomment-929869117) succinctly describes this.

Is this what Rust does? I would've expected that to at least preserve the _first_ newline. Then again, this is consistent with KDL's existing `escline` rule where `\<newline>` is the same as `<non-newline whitespace>`

CAD97:
> Is this what Rust does?

[[playground]](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d3ea961693fd48d61e674affbed2bb8c)

```rust
[src/main.rs:2] dbg!("\
    here\
    is\
    an\
    example\
    ") = "hereisanexample"
```

hkolbeck:
It's worth noting that bash behaves similarly as far as just dropping the newline, though it doesn't consume space afterward:

```
❯ echo foo\
… ❯ bar\
… ❯ baz
foobarbaz
```

With that I think `xyz` is the right output, and am +1 on including it in v2

Lucretiel:
Bash's behavior is concerned with *syntactic* whitespace (ie, allowing commands to spread over multiple lines with line continuations). It doesn't meaningfully behave in terms of consuming or not consuming specific whitespace so much as it extends a line to the next line while retaining the separation of tokens for a command. In your `echo` example, all that's happened is that the `foo` and `bar` and `baz` have correctly been passed as different arguments to `echo`; it's no different than:

```
> echo foo             bar \
   baz
foo bar baz
```

Kaydle has basically the same behavior with its own line continuation syntax, where you can use a `\` to continue a single node into the next line. All these nodes are the same:

```
node 1 2 3
node 1    2   3
node 1\
  2\
  3
```

#213 is instead concerned with treatment of escaped whitespace in strings, where I think the plain consumption of unescaped whitespace makes the most sense

> Is this what Rust does? I would've expected that to at least preserve the first newline. Then again, this is consistent with KDL's existing escline rule where \<newline> is the same as <non-newline whitespace>

Rust does just consume all whitespace, regardless of type. The canonical way to add newlines to a whitespace-escaped string to to escape them:

```rust
assert_eq!(
    "line 1\n\
    line 2\n\
    line 3\n",

"line 1
line 2
line 3
"
);
```

hkolbeck:
That makes sense, and the distinction is certainly important. Thanks for the complete writeup.

hkolbeck:
Adding escaped whitespace note to the changelog: https://github.com/kdl-org/kdl/pull/291

zkat:
Nudging the thread because I've added https://github.com/kdl-org/kdl/issues/250 to the bucket of things we should probably discuss for 2.0

hkolbeck:
Is https://github.com/kdl-org/kdl/discussions/177 worth including in discussions here?

zkat:
yeah, probably. Although I'm inclined towards having `foo` and `("")foo` be distinct values. I'm kinda iffy on the special case here.

hkolbeck:
I'm very split, on the one hand it is a special case and I'm very averse, but as @Patitotective noted this it would force implementations to distinguish between the two, which would lead to a more complex API in some languages (JS is top of mind). In addition, I just don't see a use case for blank type annotations given that impls are free to define their own.

I'm not really trying to go either direction here, just laying out thoughts.

lilyball:
Why would it be hard in JS? Can't JS just use `null` versus `""`?

I haven't really been following this and I have no experience with type annotations in KDL but my initial reaction here is that specifying `""` is potentially semantically distinct from not specifying a type annotation.

I'm curious what languages are actually expected to have a problem here? Languages without proper optional support tend to have some concept of null. Even in Go I'd expect that you could use a pointer-to-string in order to have `nil`.

That said, I've been using languages with proper optional support for long enough that I'm not sure how much of an ergonomic problem it would be to require folks to handle null type annotations in languages like Go or JS.

hkolbeck:
JS gives you null, but empty strings are falsey. It's for sure a small thing, it just means instead of writing:

```javascript
if (val.annotation) {
   ...
}
```

You have to write:

```javascript
if (val.annotation === null) {
    ...
}
```

I would count that as a more complex API.

Patitotective:
For most statically typed languages (like Nim) you have to use `Option` types which distinguish between no value and empty value. This would make APIs more complex and seems pointless to me, `("")node` should be the same as `node`.

bgotink:
I'm pro making them distinct values. Packages that work with the CST to provide an API for modifying KDL text while retaining comments and formatting would be more complex and likely inconsistent without the distinction.

If an empty or an absent annotation is considered the same, these packages would need to track that. Even if they then map empty and absent annotations onto the same public value, it would lead to confusing behaviour in the locations the different CST nodes:

```kdl
node /* comment */ ()null
//                |         < end of the leading whitespace for the `null` value
//                   |      < start of the `null` value
//                 __       < missing locations
```

Making the empty annotation `()` part of the leading whitespace would be wrong cf. the language specification.

Imo the best solution for these packages is to expose the difference between an empty and an absent annotation: consistent CST node locations and no `()` as part of whitespace.

hkolbeck:
One minor point to @bgotink's example, which I think is a good point, is the `()val` does not agree with the spec by my reading, it has to be `("")val`

CAD97:
I agree with @larsgw here in that if this is considered a problem to solve, the better solution would be to forbid zero-length identifiers rather than do some magic to make `("")` equivalent to no type annotation. Given that type annotations are not given any meaning[^1] by the specification, it's fine imho for an implementation to treat a present-but-zero-length type annotation equivalently to no type annotation.

[^1]: > KDL does not specify any restrictions on what implementations might do with these annotations. They are free to ignore them, or use them to make decisions about how to interpret a value.

So my vote here is ***no change***.

Patitotective:
I agree on considering this a problem and forbidding zero-length identifiers.
This change would convert valid tests like `blank_node_type.kdl`, `blank_arg_type.kdl`, `blanl_prop_type.kdl`, `empty_quoted_prop_key.kdl` and `empty_quoted_node_id.kdl` to invalid, it would need to be specified in the spec.
If this is okay, I'll create a PR.

Patitotective:
By the way, is there a reason why some tests use `empty` and others `blank`?

Lucretiel:
I'm pro-distinct too. I think it's fine if a particular consumer of KDL wants them to be identical, or even if a particular implementation of type hints treat them as identical, but I think that it makes sense that a KDL data model treats them as distinct (essentially, as `Option<String>`).

Lucretiel:
I'd be opposed to banning 0-length identifiers; I think that adds more complexity / confusion than it's worth. Currently, an identifier is either "bare identifier" or "quoted string", and I'm not a fan of making it instead a special non-empty subset of "quoted string".

Patitotective:
For v2.0 KDL could also allow multi-line comments not only where whitespaces can be, allowing `node (type)/*comment*/true`.

tabatkins:
We should add in the #331 change, to add vertical tab to the whitespace table.

tabatkins:
For the #204 vs #241 decision (how to handle `#` in idents), everyone who's commented in the thread seems to either be neutral or leans toward #241 (just ban idents that look rawstring-like). So I'm going to commit that into the branch.

tabatkins:
I'm strongly inclined to accept #255 and #250; they're both banning invisible control characters from idents (and for #255, matching the grammar to what the prose already indicates).

----

For #270, I'm inclined to accept it. It does make the grammar nastier, unfortunately, but being able to drop comments in wherever is quite nice when debugging. I was slightly concerned this might put us in a similar unfortunate situation as CSS where `2/**/px` was allowed (and requires [special serialiation rules](https://drafts.csswg.org/css-syntax/#serialization) to force that to serialize as `2 px` rather than `2px`, but nah, KDL doesn't go with CSS's "any token sequence is allowed, whitespace can go anywhere" philosophy. Since our whitespace allowance/requirements are strictly indicated in the grammar, this should be fine. (That is, `node/**/prop="foo"` would still be invalid, as you need whitespace between the nodename and the property name.)

-----------

For #285, inclined to accept it. It makes the grammar a little more understandable, I think, since it separates out slash-dashed things from the core notion of what forms the document. And allowing multiple slash-dashed children blocks, or interspersed slashdashed children and arguments, also seems like a reasonable thing, yeah.

----------

For #228, I'd want to do a minor editorial pass over the document before merging it, but otherwise it looks good to go.

zkat:
Hi all. I've tagged 2.0.0-draft.1, and it's ready to review! Please give it a look and let's see what else we want to change from here. If you're wondering what changed, I suggest looking at [the changelog](https://github.com/kdl-org/kdl/pull/286/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed).

Nothing is really set in stone, but I figured I would incorporate a bunch of stuff we've been sorta-kinda talking about but not really committing to, and see how it feels. I'm feeling pretty good about it!

IceDragon200:
Just some observations while updating kuddle to support the v2 spec, I may add more as I find them

## Multiline Strings

### Dedentation

> Strings may span multiple lines with literal Newlines, in which case the resulting String is "dedented" according to the line with the fewest number of Whitespace characters preceding the first non-Whitespace character. That is, the number of Whitespace characters in the least-indented line in the String body is subtracted from the Whitespace of all other lines.

> The first Newline, the last Newline, along with Whitespace following the last Newline, are not included in the value of the String. The first and last Newline can be the same character (that is, empty multi-line strings are legal).

```kdl
// So these are all valid
nl_only "
"
no_space "
ABC
"
two_spaces "
  ABC
"
// Including this, which looks funky
heredoc "
a
    "
// the outdent here is scaring me, but it's valid I suppose
heredoc2 "
  do
    some code
  end
    "

que? "
    something
      something2
        something3
  "
```

```kdl
// Since its a multiline string, but only the first newline is present, the resulting string should be empty
nl_only ""
// Multiline string with a single content  line
no_space "ABC\n"
// Multiline string with two spaces before first content line
two_spaces "  ABC\n"
// heredoc with uneven indentation, since a is the shortest line it will be used to trim the rest
// the preceding whitespace before the " is discarded
heredoc "a\n"
heredoc2 "do\n  some code\nend"

// Here is the questionable one: should it end up like this (this is what would happen in elixir)
que? "    something\n      something2\n        something3"
// Or like this (this is what would happen based on the clauses above, assuming last whitespaces are discarded)
que? "something\n  something2\n    something3"
```

So why is this section here?

It's the dedent rule that's the problem, those two clauses describe the newline and dedent behaviour, however it doesn't make clear if the whitespace AFTER the last newline is to be counted as the dedent, since it's discarded by the second clause.

### Misleading literal newlines clause

On the topic of multiline strings, I found a wording issue:

> Strings with literal Newlines that do not immediately start with a Newline and whose final " is not preceeded by whitespace and a Newline are illegal.

```kdl
node "a
"
node "
"
```

Those are both valid strings, the latter passes the `starts with a newline`, but fails the latter `preceded by whitespace and newline`.

Those would both evaluate to the following:
```
node "a\n"
node2 "\n"
```

Which is are strings with literal newlines: 
The first did not **immediately start with a newline**, and its final `"` is not **preceded by whitespace AND a newline** (keyword here AND, the final `"` is preceded by a single newline, but the wording made it illegal).

I'd argue removing that line all together, because it is misleading in its current state.

zkat:
@IceDragon200 Here's how you should interpret the following strings.

The last newline is also stripped:

```kdl
no_space "
ABC
"
no_space "ABC"
```

Things are fully dedented based on their least-indented line:
```kdl
que? "
    something
      something2
        something3
  "
que? "something\n  something2\n    something3"
```

> however it doesn't make clear if the whitespace AFTER the last newline is to be counted as the dedent, since it's discarded by the second clause.

It should not be counted. It should be fully stripped, including including the newline associated with it.

> Strings with literal Newlines that do not immediately start with a Newline and whose final " is not preceeded by whitespace and a Newline are illegal.

This could be worded better, I guess. Perhaps "Strings with literal newlines not immediates after the opening `"`, or whose final `"` is not preceded by (optional) whitespace and a newline are illegal".

So:

```kdl
// illegal
foo "a
"

// legal
foo "
"
// evaluates to:
foo ""
```

For the sake of having something to play with, these dedenting rules are based directly on 
[JavaScript's dedent proposal](https://github.com/tc39/proposal-string-dedent#proposed-solution), and there's an [interactive playground](https://output.jsbin.com/wiwovot/quiet) you can use to test out any corner cases.

If you find some inconsistency between the KDL spec and what the JS spec are describing, the KDL spec should be updated to conform to JS', since I trust them to have thought through the corner cases better than me.

IceDragon200:
> For the sake of having something to play with, these dedenting rules are based directly on
[JavaScript's dedent proposal](https://github.com/tc39/proposal-string-dedent#proposed-solution), and there's an [interactive playground](https://output.jsbin.com/wiwovot/quiet) you can use to test out any corner cases.

We need to keep in mind that proposal is intended to fix a syntactical shortcoming of JS, that is, the lack of a proper heredoc.

They even list ruby's heredoc `<<~` which does the exact thing I expected: https://github.com/tc39/proposal-string-dedent#in-other-languages

Even elixir's heredoc (which I count as the golden standard for its consistent behaviour):

```elixir
node = """
    ABC
  """

# "  ABC\n"
```

In this case, the final `"""` denotes the strings root identation level, the equivalent in KDL would be:

```kdl
node "
    ABC
  "

// Which should be this if it followed that same ruling 
node "  ABC\n"
```

Since KDL is trimming based on the content of the string, I can see it trimming actual intended indents:

```kdl
node "
            This string is intended to have its indents preserved
       Since all of the text should be centered like this line and others        
    It's an unusual test case for sure, but it's something that could happen
             The the leading whitespace on each line is intended.
           But the current implementation will end up truncating it.
"

// Would end up like this using JS dedent
node "
        This string is intended to have its indents preserved
   Since all of the text should be centered like this line and others        
It's an unusual test case for sure, but it's something that could happen
         The the leading whitespace on each line is intended.
       But the current implementation will end up truncating it.
"
```

On the topic of newline shenanigans, I'd argue as long as both quotes are present, any oddball combination should theoretically work:

```kdl
// These should resolve to the same thing
node "a
"
node "a\n"

// These are the same because of the multiline rule, trimming the immediate newline
node "
"
node ""

// These should be valid (there is a literal space after the quote followed by a newline)
// By the same rule that disallowed node "a\n", this is allowed
node " 
"
node "\s\n"

// This should be illegal, at least based on the multiline structure, but looks fine technically
node "
a"
node "\na"

// But this should be okay? (there is a space after the first quote)
node " 
a"
node "\s\na"

// Should be fine
node "\n
a"
node "\n\na"
```

I'm sure there are other oddball scenarios out there, but all of those look valid enough to me.

zkat:
I actually kinda like the idea of determining overall indentation by the indentation of the last `"`, although that means it should be a syntax error to have something like this:

```kdl
foo "
bar
  "
```

Otherwise I really don't like our alternatives for dealing with that. It should just be invalid. But communicating this constraint might feel really restrictive and confusing. That said, I understand the desire to have _the entire string_ indented.

As far as your other examples go, I think the "multi-line strings MUST be `"` + newline" is an important constraint because it clarifies that we are now in multi-line MODE, because it has special logic to it. I feel the same about having the closing `"` in its own line.

Most other languages deal with this by using special syntax for their "heredocs", but this is a way for KDL to reuse existing string syntax.

README.md
```diff
@@ -242,8 +249,8 @@ specific meanings.
 
 ```kdl
 numbers (u8)10 (i32)20 myfloat=(f32)1.5 {
-  strings (uuid)"123e4567-e89b-12d3-a456-426614174000" (date)"2021-02-03" filter=(regex)r"$\d+"
-  (author)person name="Alex"
+  strings (uuid)123e4567-e89b-12d3-a456-426614174000 (date)"2021-02-03" filter=(regex)#"$\d+"#
```
tabatkins, line 252: Need to keep quotes on the uuid, idents can't start with digits.

SPEC.md
```diff
@@ -120,26 +129,28 @@ The following characters cannot be the first character in a
...
+negative number.
 
 ### Non-identifier characters
 
 The following characters cannot be used anywhere in a [Bare Identifier](#identifier):
 
-* Any codepoint with hexadecimal value `0x20` or below.
-* Any codepoint with hexadecimal value higher than `0x10FFFF`.
-* Any of `\/(){}<>;[]=,"`
+* Any of `(){}[]/\="#;`
```
tabatkins, line 141: Oh, I thought we were just gonna disallow # as the first char, not anywhere.

SPEC.md
```diff
@@ -314,6 +324,36 @@ String is either a [Quoted String](#quoted-string) or a
...
+
+Multi-line strings _MUST_ have a single [Newline](#newline) immediately
+following their opening `"`, after which they may have any number of newlines.
+Finally, there must be a Newline, followed by any number of Whitespace, before
+the closing `"`.
+
+The first Newline, the last Newline, along with Whitespace following the last
+Newline, are not included in the value of the String. The first and last
+Newline can be the same character (that is, empty multi-line strings are
+legal).
```
tabatkins, line 349: Just checking: the following is *also* an empty string, right?

```
node "

"
```

(that is, `"\n\n"`)

It *should* be, fwiw, just verifying.

SPEC.md
```diff
@@ -314,6 +324,36 @@ String is either a [Quoted String](#quoted-string) or a
...
+
+Strings _MUST NOT_ include the code points for [disallowed literal
+code points](#disallowed-literal-code-points) directly. If needed, they can be
+specified with their corresponding `\u{}` escape.
+
+### Multi-line Strings
+
+Strings may span multiple lines with literal Newlines, in which case the
+resulting String is "dedented" according to the line with the fewest number of
+Whitespace characters preceding the first non-Whitespace character. That is,
```
tabatkins, line 337: Should make it excessively clear that only *literal* whitespace is removed. Escaped whitespace counts as content.

SPEC.md
```diff
@@ -314,6 +324,36 @@ String is either a [Quoted String](#quoted-string) or a
...
+Strings _MUST_ be represented as UTF-8 values.
+
+Strings _MUST NOT_ include the code points for [disallowed literal
+code points](#disallowed-literal-code-points) directly. If needed, they can be
+specified with their corresponding `\u{}` escape.
+
+### Multi-line Strings
+
+Strings may span multiple lines with literal Newlines, in which case the
+resulting String is "dedented" according to the line with the fewest number of
```
tabatkins, line 336: Not "fewest number", that would count a two spaces on one line the same as two tabs on the next. Should instead find the largest exactly shared whitespace prefix.

SPEC.md
```diff
@@ -93,25 +94,33 @@ foo 1 key="val" 3 {
 ### Identifier
```
tabatkins, line 94: Do we want to rephrase the ident/string concepts so idents are a type of String, rather than strings being a kind of ident? That mental change is what finally flipped me on the `#true` change.

SPEC.md
```diff
@@ -334,32 +379,79 @@ interpreted as described in the following table:
...
+
+##### Escaped Whitespace
+
+In addition to escaping individual characters, `\` can also escape whitespace.
+When a `\` is followed by one or more literal whitespace characters, the `\`
+and all of that whitespace are discarded. For example, `"Hello World"` and
+`"Hello \    World"` are semantically identical. See [whitespace](#whitespace)
+and [newlines](#newlines) for how whitespace is defined.
+
+Note that only literal whitespace is escaped; *escaped* whitespace is retained.
```
tabatkins, line 396: We should change the phrasing a bit here - this section is *about* "escaped whitespace", but you're talking about "escaped whitespace" as being something different here. I suggest "whitespace escape" for `\n` and the like.

tabatkins: Or maybe "discarded whitespace" for this section, instead.

SPEC.md
```diff
@@ -334,32 +379,79 @@ interpreted as described in the following table:
...
+           bar
+  "#
+```
+
+The last example's string value will be:
+
+```
+    foo
+This is the base indentation
+        bar
```
tabatkins, line 454:
````suggestion
just-escapes #"\n will be literal"#
```
The string contains the literal characters `\n will be literal`.

```kdl
quotes-and-escapes ##"hello\n\r\asd"#world"##
```
The string contains the literal characters `hello\n\r\asd"#world`

```kdl
multi-line #"
        foo
    This is the base indentation
           bar
  "#
```

The last example's string value will be:

```
    foo
This is the base indentation
        bar
````

File: SPEC.md [hunk 10]
```diff
@@ -334,32 +379,79 @@ interpreted as described in the following table:
 | Carriage Return               | `\r`   | `U+000D` |
 | Character Tabulation (Tab)    | `\t`   | `U+0009` |
 | Reverse Solidus (Backslash)   | `\\`   | `U+005C` |
-| Solidus (Forwardslash)        | `\/`   | `U+002F` |
 | Quotation Mark (Double Quote) | `\"`   | `U+0022` |
 | Backspace                     | `\b`   | `U+0008` |
 | Form Feed                     | `\f`   | `U+000C` |
-| Unicode Escape                | `\u{(1-6 hex chars)}` | Code point described by hex characters, up to `10FFFF` |
+| Unicode Escape                | `\u{(1-6 hex chars)}` | Code point described by hex characters, as long as it represents a [Unicode Scalar Value](https://unicode.org/glossary/#unicode_scalar_value) |
+| Whitespace Escape             | See below | N/A   |
```
tabatkins, line 386: To help with marking a dedent, do we want to add an escape code for a normal space? I suggest just `\s`. That way you can escape the first significant space easily, without having to remember that it's U+0020 (and `\s` is easier to write than `\u{20}`).

SPEC.md
IceDragon200: If it was left it the raw string form would be misinterpreted, from what I gathered from conversations is # is going ot be used for keywords (only `#true`, `#false` and `#null` atm)

SPEC.md
IceDragon200: See our discussion above about these cases

SPEC.md
zkat: It originally was, but then I thought it might be too confusing to _only_ ban it as the first character. I'm open to banning it only as an initial character, though. Any strong feelings?

SPEC.md
zkat: hmmmm. You're right. Based on what I wrote, this would be an empty string, but I would expect that node to be equivalent to `node "\n"`.

SPEC.md
IceDragon200: Oh crap, I didn't even realize it wasn't present in the table either, I added ALL the usual escapes (\s\r\n\f\b)

https://github.com/IceDragon200/kuddle/pull/1/files#diff-1663e36dafaff337ff4883c20f375c3cebf1de34e44f592012ddb870e8e29651R196-R214

Maybe we should add it, it doesn't really hurt anything

SPEC.md
zkat: this is actually the intention: for two tabs to be treated the same as two spaces, as far as indentation goes. I'm not sure we should be making any decisions about how many spaces wide a tab is?

IceDragon200: Actually that brings up a good point, what happens if someone mixes spaces and tabs into a multiline string? 

How do you even dedent that?

SPEC.md
zkat: I like this idea. It may be a way for us to work around the issue @IceDragon200 brought up about dedenting text that should have _some_ global indentation (the suggested alternative being to determine dedent level based on the closing `"`'s indentation, like Elixir does).

SPEC.md
zkat: like the text says: based on the number of characters. `\t\t` and `\s\s` count as two, and both mean that you remove "two items of whitespace" from the other lines, regardless of whether they're tabs or spaces.

SPEC.md
zkat: Oh, actually, the JS spec implementation also treats that as `""`, not `"\n"`. Interesting.

SPEC.md
tabatkins: @IceDragon200 No, `foo#bar` is not ambiguous with anything.

@zkat I don't have *strong* feelings about it, but I'm weakly in favor of allowing it in non-initial positions. Whichever way you prefer is fine, tho.

SPEC.md
tabatkins: Right, the following

```
node "
	foo
	"
```

should be equivalent to `node "foo"`, so my example should similarly just be the empty string. I was just double-checking and wondering if we wanted to make that a little clearer.

SPEC.md
tabatkins: The JS proposal did a survey of langs; several *do* just count the characters (and one treats tabs as equivalent to a number of spaces), and the champions explicitly didn't like those and went with exact prefix match. I also prefer that; it's weird that the following will dedent both lines to have zero whitespace:

```
node "
    spaced
				tabbed
"
```

> I'm not sure we should be making any decisions about how many spaces wide a tab is?

Correct, we shouldn't; exact ws prefix match only. So my example above will *not dedent at all*, since the common ws prefix is nothing.

People shouldn't be mixing tabs and spaces between lines, but if they do, I *guarantee* they don't expect one tab to equal one space for removal. Dedenting is meant to achieve a *visual* goal.

zkat: @tabatkins what do you think of doing what @IceDragon200 suggested [here](https://github.com/kdl-org/kdl/pull/286#issuecomment-1854704718), and, instead of looking for "common prefix", doing "strip whitespace matching the whitespace between the final newline and the closing `"`"?

That would allow literal multiline strings to be themselves fully-indented, and give a clear, single point that users could adjust to set the desired indentation of their strings, without weird `\s` trickery?

SPEC.md
zkat: I think I really like this idea, and I was definitely feeling something along these lines might help it be more clear that they're just a different type of string as far as the language is concerned.

SPEC.md
zkat: I think I'm going to keep it like this. Tbh if we allow `#` as a non-initial character, we may as well also allow `"/\`, possibly even the bracket types, since once we know an identifier doesn't start with one of these characters, we can unambiguously parse the rest of the ident until the next whitespace. Even for `{`, which now requires a space before it. The question is: at what point does all this "you can use these characters, but _not as first characters_ become too much cognitive load for users? Maybe it's best that we make all our structure-significant tokens completely banned from identifiers?

It also protects about silly mistakes like:

```kdl
node foo#true
```

Thoughts?

Bannerets:
> - All strings in KDL are multi-line, and raw strings are written with Rust-style syntax (`r"foo"`), instead of backticks.

Looks like this line in the README should also be changed to use the new raw string syntax.

SPEC.md
zkat: this is done now

SPEC.md
zkat: This is node!

IceDragon200:
Test case `unusual_bare_id_chars_in_quoted_id.kdl` needs updating as it still contains a `#` which is considered illegal

IceDragon200:
Unicode test for RLM and LRM are present, but the specific characters are not mentioned in the linked reference: https://www.w3.org/International/questions/qa-bidi-unicode-controls, and not references in the spec itself.

tl;dr We need to add `U+200F` and `U+200E` to the list of disallowed characters

IceDragon200:
## Invalid/Incorrect Tests

* `all_escapes.kdl` (`\s` does not need to be escaped on output)
* `null_arg.kdl` (incorrect output, input was `#null` keyword, output is currently `null` bare identifier, should be `#null`)
* `escline.kdl` (output arg does not need to be quoted)
* `empty_quoted_prop_key.kdl` (output "empty" does not need to be quoted)

## Newlines

Some expected tests have extraneous newlines or not enough newlines at the end of the file:

* `arg_bare.kdl` (missing newline)
* `prop_identifier_type.kdl` (extraneous newline)
* `question_mark_before_number.kdl` (missing newline)
* `dash_dash.kdl` (missing newline)

zkat:
@IceDragon200 LRM/RLM are listed [in a different section](https://www.w3.org/International/questions/qa-bidi-unicode-controls#rlmlrm), and there were already tests for it. I've added them to the spec, though, that was an oversight.

Re: incorrect tests, thanks for the review! I'll go fix all these now!

SPEC.md
tabatkins: Yeah I'm fine with that. I lean *slightly* towards matching JS `dedent` just because it makes reasonable decisions overall here and matching precedent is nice, but if you prefer "last line sets the indent, and it's a parse error for any other non-empty line to not have an exactly matching prefix", I'm okay with that.

SPEC.md
tabatkins: Yeah I'm fine with this. Your general point is valid (tho we definitely shouldn't ban the numeric characters inside of idents, but I doubt you were intending to refer to those).

zkat: yeah I'm ok with this change as-is. Thanks

SPEC.md
zkat: The only thing that convinces me to do last-line instead of common-prefix is the ability to easily control overall-final-indent, which seems valuable. So I think I'll go with that...

IceDragon200:
## Tests with discussions

* `type_before_prop_key.kdl` https://github.com/kdl-org/kdl/discussions/362

## Mismatch Tests

* `unusual_bare_id_chars_in_quoted_id.kdl` Input / Output mismatch, input contains quoted string with `#`, output is bare identifier with `#`

## Impossible Tests?

* `unicode_scalar_low.kdl` and `unicode_scalar_high.kdl` the input file(s) are already incorrect, the character present there is U+FFFE. It is almost impossible to embed 0xD800..0xDFFF as most self respecting languages won't even allow you to generate the unicode for them to begin with
 
## Old Keyword is now bare identifier

The following tests are now all valid, as the identifiers here are no longer reserved

* `null_prop_key.kdl`
* `true_prop_key.kdl` 
* `false_prop_key.kdl`

## Multiline String Tests

* `slashdash_full_node.kdl` the string at the end of the test is invalid according to the multiline string rule, rendering the entire test invalid
* `escaped_whitespace.kdl` due to multiline string rules, one of the strings are considered invalid:
```kdl
"Hello
	World" \
```
* `multiline_string.kdl` output is incorrect, there are no leading or tailing newlines (multi-line string rules)
* `multiline_raw_string.kdl` same as above (multi-line string rules)
* `raw_string_newline.kdl` isn't this the same test as above?

## Oddities with node /wo children

These tests drop their brackets on output, not sure if that's intended

* `empty_child_same_line.kdl`
* `empty_child_different_lines.kdl`
* `empty_child_whitespace.kdl` 

It might just be kuddle, but the presence of the brackets denote the possibility of children, vs without [^1].
```kdl
// Implicit
node
// Explicit
node { }
```

## Should this be valid?

* `comment_after_arg_type.kdl` At a glance, this should be valid I think:
```elixir
# Just some pseudo code on what kuddle would tokenize this to
[{:term, "node"}, :space, :(, {:term, "type"}, :), {:comment, :ml, "huh"}, {:term, "10"}] #=> node (type)/*huh*/10

# The comment would be dropped during parsing
[{:term, "node"}, :space, :(, {:term, "type"}, :), {:term, "10"}] #=> node (type)10
```
 
[^1]: In respect to kuddle, nodes with or without brackets are maintained, there is a difference between a node with no implicit children (`children: nil`) and a node with no explicit children (`children: []`)

SPEC.md
```diff
@@ -455,34 +590,71 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
 
 Note that for the purpose of new lines, CRLF is considered _a single newline_.
 
+### Disallowed Literal Code Points
+
+The following code points may not appear literally anywhere in the document.
+They may be represented in Strings (but not Raw Strings) using `\u{}`.
+
+* Any codepoint with hexadecimal value `0x20` or below (various control
```
Bannerets, line 598: 0x20 is space, so this seems to imply that space is not allowed inside quoted strings.

Also, should BOM be included here? (Otherwise, I guess, there could be a bit of conflict of how `<U+FEFF>node` at the beginning of the document should be parsed. Disambiguating by longest token, it is parsed as one identifier.)

SPEC.md
```diff
@@ -302,17 +273,71 @@ IEEE 754-2008 decimal floating point numbers
...
 
-Strings in KDL represent textual [Values](#value), or unusual identifiers. A
-String is either a [Quoted String](#quoted-string) or a
-[Raw String](#raw-string). Quoted Strings may include escaped characters, while
-Raw Strings always contain only the literal characters that are present.
+Strings in KDL represent textual UTF-8 [Values](#value). A String is either an
+[Identifier String](#identifier-string), a [Quoted String](#quoted-string) or
+a [Raw String](#raw-string). Quoted Strings may include escaped characters,
+while Raw Strings always contain only the literal characters that are present.
+Identifier Strings don't user delimiters.
```
Bannerets, line 287: sounds like a typo
```suggestion
Identifier Strings don't use delimiters.
```

SPEC.md
bgotink: The "0x20 or below rule" also includes `0x09` (the tab `\t`), `0x0A` (newline `\n`), `\x0B` (vertical tab `\v`), `0x0C` (form feed `\f`), and `0x0D` (carriage return `\r`).

These are all considered either newline or whitespace in this spec and should be excluded from this rule.

tabatkins: fixed

SPEC.md
tabatkins: I've gone ahead and edited the PR (and associated tests) to match this; the final line now gives the whitespace to remove. We'll want a bunch more tests for this, but #361 is already covering that.

tabatkins:
@IceDragon200 Re your comment about several tests

* Mismatch tests: Fixed. Removed the `#` from both of them, since they're intended to demonstrate serializing an unusual ident string.
* Impossible Tests: Correct, these are both wrong as written and actually impossible, given that KDL docs must be UTF-8 in the first place. Removed.
* Old keyword is now bare identifier: Nope, those are still properly invalid; we reserved the identifiers "null", "true", and "false" from being ident strings.
* Multiline string tests: all fixed
* Oddities with node w/o children: This has been discussed in the past; currently KDL doesn't require you to preserve the distinction between "no child list" and "empty child list". 
* Should this be valid? Yes it's valid now, there's an `optional-node-space` between a value's tag and main body. Fixed (all three of those tests, plus the similar `comment_in_...` ones).

IceDragon200:
@tabatkins One more issue, how is the dot_zero (`.0`) being restricted:

```
identifier := string | bare-identifier
bare-identifier := (unambiguous-ident - boolean - 'null') | numberish-ident
unambiguous-ident := (identifier-char - digit - sign) identifier-char*
numberish-ident := sign ((identifier-char - digit) identifier-char*)?
identifier-char := unicode - line-space - [\\/(){};\[\]="#] - disallowed-literal-code-points

sign := '+' | '-'
```

Because looking over that, `.0` is legal (it falls under unambiguous-ident 
```
[{:identifier-char, "."}, {:digit, "0"}] 
```

tests/test_cases/input/eof_after_escape.kdl
```diff
@@ -0,0 +1 @@
+node \
```
IceDragon200, line 1: This is a valid test, escline can be followed immediately by eof:

```
escline := '\\' ws* (single-line-comment | newline | eof)
```

tabatkins:
Per #366, I've made `.1`/etc illegal as idents. I've also done some minor editorial rearranging of the grammar so that it matches the spec concepts a little better.

IceDragon200:
@tabatkins Any tests for the new .0 restrictions (including prefixes)?

tabatkins:
I didn't add any, but it'll definitely need some. One sec.

zkat:
I just tagged another draft of the spec.

I think the biggest pending item for the new spec is addressing the way we do multiline string dedenting. I think we've all agreed that the way that's currently described is not what we want. So the question is do we want to do what JS does ("common indentation"), or by the indentation of the closing `"`.

Any strong opinions? I'm personally leaning towards the latter, since it would allow multiline strings which are themselves indented, rather than always stripping indentation even when you might want it (which JS would do).

tabatkins:
Well, in JS you can keep indentation at a particular level by explicitly escaping the first whitespace you want to preserve, on one of your lines. That is, 

```js
var oneTabInIt = String.dedent`
	\tfoo
		bar
		baz
	`;
```

But that does still look awkward, I'll admit. I think "dedent according to the final line" is reasonable.

So, just to be clear:
* indentation must be an exact whitespace match (same sequence of spaces, tabs, etc), not just "same count of whitespace chars"
* incorrect indenting is a syntax error

Right?

IceDragon200:
I'm in favour of final quote designating the multiline spaces, it's a straightforward rule to understand and removes the guess work for the most part (most modern editors highlight indentation columns for example, making it easy to quickly scan and tell how deep an indentation will actually be when parsed)

tabatkins:
Screenshot from 2024-02-06 15-30-39

zkat:
@tabatkins that sounds good to me, yes. I like it being a clear syntax error.

zkat:
Alright, draft 3 is out.

I think this is a good time to start recommending that existing implementations start working on support for 2.0. Any thoughts? I don't think there's anything missing at this point, except maybe we can add #371 in a bit once it's hashed out a bit (but that's just for the schema spec).

Thanks for all the work so far, everyone! We're almost at the finish line! This has been in the works for like... 2 years? 1.5 years?

IceDragon200:
[Kuddle](https://github.com/IceDragon200/kuddle/pull/1) has been updated, I was just waiting for the spec to finalize before merging it into main/master, it's also what I've been using to test the spec essentially.

Do we have an issue tracking v2 implementations yet? It would give everyone a good idea of what's available for testing, maybe the community can find bugs we didn't notice and so forth

zkat:
@IceDragon200 https://github.com/kdl-org/kdl/issues/372 this is the epic for tracking implementation support

zkat:
I was trying to think of what sort of changes this would entail as far as consumer impact, and I was pleasantly surprised that _all_ of our changes seem to be syntactic: the data model didn't change at all, so anyone who uses an existing KDL library should be able to use the exact same API calls in the same way, except the parsers would be different.

SPEC.md
```diff
@@ -413,6 +605,7 @@ space](https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt):
 | Name                 | Code Pt |
 |----------------------|---------|
 | Character Tabulation | `U+0009`  |
+| Line Tabulation      | `U+000B`  |
```
larsgw, line 608: This has also been added as a Disallowed Literal Code Point right?

larsgw: And U+0009 too.

zkat: wait, this seems like an error. Why did we disallow literal tabs?...

larsgw: 491cc46f89df0261b494ccc89c2bc7b243467bb2

tabatkins:
Hm, I did it in response to <https://github.com/kdl-org/kdl/pull/286#discussion_r1433274645>, but clearly I fucked up that edit *hardcore*.

SPEC.md
tabatkins: Well, I guess I just got the range wrong. It should be 0x00-0x08. The rest are fine, I think.

larsgw: 0x0B should be removed from the range too, I think.

SPEC.md
```diff
@@ -455,34 +670,72 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
+
+* The codepoints `U+0000-0009`,
+    the codepoint `U+000B`,
+    or the codepoints `U+000E-001F`  (various control characters).
+* `U+007F` (the Delete control character).
+* Any codepoint that is not a [Unicode Scalar
+  Value](https://unicode.org/glossary/#unicode_scalar_value).
+* `U+2066-2069`, `U+202A-202E`, `U+200E`, and `U+200F`, the [unicode
+  "direction control"
+  characters](https://www.w3.org/International/questions/qa-bidi-unicode-controls)
```
larsgw, line 686: While we're at it, let's put these in order.

SPEC.md
```diff
@@ -455,34 +670,72 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
+
+The following code points may not appear literally anywhere in the document.
+They may be represented in Strings (but not Raw Strings) using `\u{}`.
+
+* The codepoints `U+0000-0009`,
+    the codepoint `U+000B`,
+    or the codepoints `U+000E-001F`  (various control characters).
+* `U+007F` (the Delete control character).
+* Any codepoint that is not a [Unicode Scalar
+  Value](https://unicode.org/glossary/#unicode_scalar_value).
```
larsgw, line 683: This is U+D800 to U+DFFF (inclusive) then? Can we include that in the spec for convenience?

SPEC.md
```diff
@@ -455,34 +670,72 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
+type := '(' optional-node-space string optional-node-space ')'
+equals-sign := See Table (Equals Sign)
+
+string := identifier-string | quoted-string | raw-string
+
+identifier-string := unambiguous-ident | signed-ident | dotted-ident
+unambiguous-ident := ((identifier-char - digit - sign - '.') identifier-char*) - 'true' - 'false' - 'null'
+signed-ident := sign ((identifier-char - digit - '.') identifier-char*)?
+dotted-ident := sign? '.' ((identifier-char - digit) identifier-char*)?
+identifier-char := unicode - line-space - [\\/(){};\[\]="#] - disallowed-literal-code-points
```
larsgw, line 726: `line-space` is not a character class anymore, should `- line-space` be `- newline - unicode-space`?

larsgw: The new equal signs also still need to be added (i.e. excluded).

SPEC.md
```diff
@@ -322,9 +346,15 @@ string characters except unescaped `"` and `\`. This includes literal
 multiple lines without behaving like a Newline for [Node](#node) parsing
 purposes.
```
larsgw, line 347: I think this sections needs to be reworded: according to the grammar, the string values can no longer "just" have literal newline characters, it needs to be a valid multi-line string. (In fact, literal newline characters are regulated pretty heavily now.)

SPEC.md
```diff
@@ -496,21 +748,47 @@ hex := sign? '0x' hex-digit (hex-digit | '_')*
...
 
-ws := bom | unicode-space | multi-line-comment
+ws := unicode-space | multi-line-comment
 
 bom := '\u{FEFF}'
 
-unicode-space := See Table (All White_Space unicode characters which are not `newline`)
+disallowed-literal-code-points := See Table (Disallowed Literal Code Points)
+
+equals-sign := See Table ([Equals Sign](#equals-sign))
```
larsgw, line 765: Not the biggest issue but this rule is duplicated now, the other one is under the prop/value/type rules.

CHANGELOG.md
```diff
@@ -0,0 +1,74 @@
...
+  characters).
+* `,`, `<`, and `>` are now legal identifier characters. They were previously
+  reserved for KQL but this is no longer necessary.
+* Code points under `0x20`, code points above `0x10FFFF`, Delete control
+  character (`0x7F`), and the [unicode "direction control"
+  characters](https://www.w3.org/International/questions/qa-bidi-unicode-controls)
+  are now completely banned from appearing literally in KDL documents. They
+  can now only be represented in regular strings, and there's no facilities to
+  represent them in raw strings. This should be considered a security
+  improvement.
```
larsgw, line 24: "Code points under `0x20`" → except newline and whitespace code points.

zkat:
Alright, I've flipped this PR to "ready to review". Time to see what kinds of comments we get from the larger community :)

SPEC.md
```diff
@@ -334,32 +365,202 @@ interpreted as described in the following table:
...
+When a Quoted or Raw String spans multiple lines with literal, non-escaped
+Newlines, it follows a special multi-line syntax that automatically "dedents"
+the string, allowing its value to be indented to a visually matching level if
+desired.
+
+A Multi-line string _MUST_ start with a [Newline](#newline) immediately
+following its opening `"`. Its final line _MUST_ contain only whitespace,
+followed by a single closing `"`. All in-between lines that contain
+non-whitespace characters _MUST_ start with the exact same whitespace as the
+final line (precisely matching codepoints, not merely counting characters).
```
Lucretiel, line 449: Just to be clear, this means that the string *must* include a trailing empty line with whitspace? These are all illegal:

```
     "
    Hello
    World"

    "
    Hello
    World
"
```

SPEC.md
```diff
@@ -334,32 +365,202 @@ interpreted as described in the following table:
...
+### Multi-line Strings
+
+When a Quoted or Raw String spans multiple lines with literal, non-escaped
+Newlines, it follows a special multi-line syntax that automatically "dedents"
+the string, allowing its value to be indented to a visually matching level if
+desired.
+
+A Multi-line string _MUST_ start with a [Newline](#newline) immediately
+following its opening `"`. Its final line _MUST_ contain only whitespace,
+followed by a single closing `"`. All in-between lines that contain
```
Lucretiel, line 447: Unless I missed it, I don't see an elaboration on this point about "lines that contain non-whitespace characters must include the whitespace prefix. Presumably that means you're allowed to have newline pairs (for instance, as paragraph breaks) that don't have to include the whitespace prefix.

It's not totally clear to me how this is handled; my assumption would be that any line containing only literal whitespace (0 or more) would be replaced with just the newline.

SPEC.md
larsgw: According to the grammar (and #320), the first one should be illegal, the second one should be fine.

SPEC.md
```diff
@@ -334,32 +365,202 @@ interpreted as described in the following table:
...
+non-whitespace characters _MUST_ start with the exact same whitespace as the
+final line (precisely matching codepoints, not merely counting characters).
+
+The value of the Multi-line String omits the first and last Newline, the
+Whitespace of the last line, and the matching Whitespace prefix on all
+intermediate lines. The first and last Newline can be the same character (that
+is, empty multi-line strings are legal).
+
+Strings with literal Newlines that do not immediately start with a Newline and
```
Lucretiel, line 456: Should this be non-escaped literal newlines?

SPEC.md
Lucretiel: Update: I see that this point is made down in the examples. Might be worth including here in the rules.

SPEC.md
```diff
@@ -455,34 +683,72 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
+
+string := identifier-string | quoted-string | raw-string
+
+identifier-string := unambiguous-ident | signed-ident | dotted-ident
+unambiguous-ident := ((identifier-char - digit - sign - '.') identifier-char*) - 'true' - 'false' - 'null'
+signed-ident := sign ((identifier-char - digit - '.') identifier-char*)?
+dotted-ident := sign? '.' ((identifier-char - digit) identifier-char*)?
+identifier-char := unicode - unicode-space - newline - [\\/(){};\[\]"#] - disallowed-literal-code-points - equals-sign
+
+quoted-string := '"' (single-line-string-body | newline multi-line-string-body newline ws*) '"'
```
larsgw, line 741: `ws` should be `unicode-space`

SPEC.md
```diff
@@ -455,34 +683,72 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
+multi-line-string-body := string-character*
+string-character := '\' escape | [^\\"] - disallowed-literal-code-points
+escape := ["\\bfnrt] | 'u{' hex-digit{1, 6} '}' | (unicode-space | newline)+
 hex-digit := [0-9a-fA-F]
 
-raw-string := 'r' raw-string-hash
-raw-string-hash := '#' raw-string-hash '#' | raw-string-quotes
-raw-string-quotes := '"' .* '"'
+raw-string := '#' raw-string-quotes '#' | '#' raw-string '#'
+raw-string-quotes := '"' (single-line-raw-string-body | newline multi-line-raw-string-body newline ws*) '"'
```
larsgw, line 749: `ws` should be `unicode-space`

SPEC.md
zkat: oh yes.

So:

```kdl
foo "\
string"
```

should be a "single-line" string, yes?

zkat: actually do we even allow this? Do we want to?...

QUERY-SPEC.md
```diff
@@ -129,14 +102,20 @@ Then the following queries are valid:
...
+selector-operator := ">>" | ">" | "++" | "+"
+filter := matcher+
+matcher := "top()"| "()" | identifier | type | accessor-matcher
+accessor-matcher := "[" (comparison | accessor)? "]"
+comparison := accessor q-ws* matcher-operator q-ws* (type | identifier | string | number | keyword)
+accessor := "val(" number ")" | "prop(" identifier ")" | "name()" | "tag()" | "values()" | "props()" | identifier
+matcher-operator := "=" | "!=" | ">" | "<" | ">=" | "<=" | "^=" | "$=" | "*="
+
+q-ws := bom | unicode-space
+```
```
larsgw, line 121: - `identifier` should be `string`
- `val()` now also accepts decimal floats, as well as hex/octal/binary (I'm not opposed to keeping the latter)
- For consistency we might want to allow `q-ws` in the accessor and matcher brackets, at least those that take an "argument"
- `filter` now allows multiple immediately sequential `identifier`s (i.e. `string`s). Depending on how the grammar is defined with respect to backtracking etc., this would make `true` a valid query, parsed as two sequential identifiers, e.g. `tru` and `e`.

SPEC.md
larsgw: It is allowed in the grammar, it seems.

tabatkins: We don't currently allow that; if you have a multiline, the closing quote must be on a line of its own, with only the dedent space accompanying it.

I don't think we want to allow it; it complicates the model a bit (right now, it's clear that the string content is strictly the lines between the first and last), and since it wouldn't interact with dedent, your string is by necessity already starting from the left edge of your document, so having the closing quote down there as well is fine, visually.

SPEC.md
tabatkins: Yeah, we need to make sure that an *empty* line is allowed, even if the rest of the lines are indented, but if there *is* content on the line (even just whitespace), it should follow all the normal rules - it must start with the identical whitespace as the final line, and that whitespace is stripped from the string's value.

SPEC.md
tabatkins: Yeah, it's allowed, and I think it's acceptable, if weird. A string is considered "multi-line", per the spec, when it contains one or more *non-escaped* newlines.

QUERY-SPEC.md
zkat: oops. Can you take a look at the new grammar and tell me what you think?

SPEC.md
zkat: Oh yes. No problems, then.

QUERY-SPEC.md
larsgw: I think it's good to fix the order like this a bit, but we should probably mention it in the text as well. `top()` doesn't really make sense combined with other matchers (since it's not really a node) so I'm not sure I'd allow that. Type matchers are not optional at the moment, but if we fix that we need to ensure at least one matcher is used. Something like this?

```
filter := "top(" q-ws* ")" | matchers
matchers := type-matcher string? accessor-matcher* | string accessor-matcher* | accessor-matcher+
type-matcher := "(" q-ws* ")" | type
```

We can replace `q-ws` with `unicode-space` now that `bom` has been moved, but I am fine with keeping it for brevity. If we'd like `/**/` comments in KQL we could also use `ws`.

zkat: fix'd

zkat:
Alright, draft 4 is out, with #inf/#-inf/#nan support and a grammar fix!

danini-the-panini:
Should the `multiline_string_indented` test case have an extra newline in the input and the expected?

danini-the-panini:
Same for `multiline_raw_string_indented`, the expected file has an extra newline

danini-the-panini:
Should ESCLINE we allowed between `=` ?
e.g.
```
node foo=\
  "bar"
```
```
node foo\
  ="bar"
```
```
node foo=\ // lorem ipsum
  "bar"
```
etc.

SPEC.md
```diff
@@ -455,36 +696,74 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
 
+This is the full official grammar for KDL and should be considered
+authoritative if something seems to disagree with the text above. The [grammar
+language syntax](#grammar-language) is defined below.
+
 ```
-nodes := linespace* (node nodes?)? linespace*
+document := bom? nodes
+
+nodes := ('/-' plain-node-space* node)? (line-space* node)* line-space*
```
bgotink, line 722: If I understand this correctly, the following is valid KDL

```kdl
one
two
three
```

but this isn't:

```kdl
one
/-two
three
```

because `line-space` requires at least one `plain-line-space` before any slash-dashed node and the newline after node `one` is already consumed as the `node-terminator`.

Am I missing something or is there something wrong here?

tests/test_cases/expected_kdl/multiline_string_indented.kdl
```diff
@@ -0,0 +1,2 @@
+node "  hey\n everyone\n   how goes?"
+
```
danini-the-panini, line 2:
```suggestion
```

tests/test_cases/expected_kdl/multiline_raw_string_indented.kdl
```diff
@@ -0,0 +1,2 @@
+node "  hey\n everyone\n   how goes?"
+
```
danini-the-panini, line 2:
```suggestion
```

SPEC.md
zkat: Is that right? I'm not sure how to resolve it off the top of my head. I could use some help 😅

Lucretiel: Doesn't the `*` operator resolve this? You need 0-or-more additional linespace. The bigger problem I'm seeing is that there's no repetition that I can see for slashdash nodes.

bgotink: @Lucretiel `line-space` contains the slashdash:

```
line-space := plain-line-space+ ('/-' plain-node-space* node)?
```

I think the slashdash in `nodes` was added to allow documents to start with a slashdash? 

@zkat  Could we define `line-space` as follows?

```
line-space := plain-line-space+ | '/-' plain-node-space* node
```

`node` already forces a `node-terminator` at the end, so I don't think a `plain-line-space` is necessary before a slash-dashed node: `one;/-two;three` should be valid, just like `one;two;three` already is.

(This change would also allow removing the `('/-' plain-node-space* node)?` at the start of `nodes`)

JSON-IN-KDL.md
bgotink, line 1: Should the JiK spec mention that `#inf`, `#-inf`, and `#nan` are invalid values in a JiK context?
It is already implied, as JSON can't represent these values, but I think it would make sense to make this explicit.

tabatkins: Good for a warning, yeah.

CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+* Raw strings no longer require an `r` prefix: they are now specified by using
+  `#""#`.
+* Line continuations can be followed by an EOF now, instead of requiring a
+  newline (or comment). `node \<EOF>` is now a legal KDL document.
+* `#` is no longer a legal identifier character.
+* `null`, `true`, and `false` are now `#null`, `#true`, and `#false`. Using
+  the unprefixed versions of these values is a syntax error.
```
chitoyuu, line 32: I'm sorry if this has been explained somewhere before, a 100+ comment discussion can be hard to follow.

Speaking purely from a user's perspective, this specific change feels a bit unnecessary. I assume it's being done to prevent ambiguity, but `null`, `true` and `false` are keywords common enough, that I don't think anything with the slightest experience writing code would be surprised that they have special meanings unlike normal identifiers. If anything, being a Rust user, I would come into this expecting the exact opposite: that `true` means the boolean and `#true` means the raw identifier, similar to how it works in Rust sans the `r`.

Is there something I'm missing here?

zkat: We decided this would be enough of a potential footgun since the change to allow unquoted strings. Programming languages like Rust have other defenses against this kind of confusion, but kdl would need to do something different (like this prefixing) to prevent, say, the kind of things you see happen in plain JavaScript

chitoyuu: Thanks for the explanation, and fair enough, if that's the point of balance you've decided on for your project.

README.md
```diff
@@ -141,36 +141,67 @@ node1; node2; node3;
...
 
 ```kdl
-string "my
-multiline
-value"
+exec #"
+  echo "foo"
+  echo "bar"
+  cd C:\path\to\dir
+"#
```
Bannerets, line 190:
```suggestion
  "#
```
Similarly in this example

examples/ci.kdl
```diff
@@ -1,47 +1,52 @@
...
       }
-      step "Clippy" run="cargo clippy --all -- -D warnings"
-      step "Run tests" run="cargo test --all --verbose"
+      step Clippy { run cargo clippy --all -- -D warnings }
+      step "Run tests" { run cargo test --all --verbose }
+      step "Other Stuff" run="
+        echo foo
+        echo bar
+        echo baz
+      "
```
Bannerets, line 49:
```suggestion
        "
```
Again

SPEC.md
```diff
@@ -455,36 +696,74 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
...
+unambiguous-ident := ((identifier-char - digit - sign - '.') identifier-char*) - 'true' - 'false' - 'null' - 'inf' - '-inf' - 'nan'
+signed-ident := sign ((identifier-char - digit - '.') identifier-char*)?
+dotted-ident := sign? '.' ((identifier-char - digit) identifier-char*)?
+identifier-char := unicode - unicode-space - newline - [\\/(){};\[\]"#] - disallowed-literal-code-points - equals-sign
+
+quoted-string := '"' (single-line-string-body | newline multi-line-string-body newline unicode-space*) '"'
+single-line-string-body := (string-character - newline)*
+multi-line-string-body := string-character*
+string-character := '\' escape | [^\\"] - disallowed-literal-code-points
+escape := ["\\bfnrt] | 'u{' hex-digit{1, 6} '}' | (unicode-space | newline)+
```
Bannerets, line 758: The `s` escape is missing here

SPEC.md
```diff
@@ -455,36 +696,74 @@ lines](https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf):
 
 Note that for the purpose of new lines, CRLF is considered _a single newline_.
 
+### Disallowed Literal Code Points
```
Bannerets, line 699: Should BOM/ZWNBSP be included in this list? The changelog states that BOM is now prohibited except as the first character, but that is not reflected anywhere in the spec (which should imply that ZWNBSP is allowed in an identifier).

Bannerets:
How exactly multi-line strings interact with whitespace escapes (`\  `)? Perhaps that should be clarified in the SPEC?

That is, are these allowed?

```
  "
  foo \
 bar
  baz
  "
```

```
  "
  foo
  bar\
  "
```

SPEC.md
```diff
@@ -334,32 +365,202 @@ interpreted as described in the following table:
...
+| Unicode Escape                | `\u{(1-6 hex chars)}` | Code point described by hex characters, as long as it represents a [Unicode Scalar Value](https://unicode.org/glossary/#unicode_scalar_value) |
+| Whitespace Escape             | See below | N/A   |
+
+##### Escaped Whitespace
+
+In addition to escaping individual characters, `\` can also escape whitespace.
+When a `\` is followed by one or more literal whitespace characters, the `\`
+and all of that whitespace are discarded. For example, `"Hello World"` and
+`"Hello \    World"` are semantically identical. See [whitespace](#whitespace)
+and [newlines](#newlines) for how whitespace is defined.
```
Bannerets, line 381:
```suggestion
and [newlines](#newline) for how whitespace is defined.
```

danini-the-panini (approved):

SPEC.md
zkat: Yeah, let's do that.

zkat:
@Bannerets I've made some changes to clarify this. wdyt?

zkat:
🤔 maybe it should be the other way around: it might be easier to implement to have escapes process _first_, and _then_ do multiline stuff. It might also make more sense, too.

But if you're using multiline escapes in multiline strings, you're asking for trouble anyway.

Bannerets:
I think interpreting escapes after dedenting seems intuitive, in the sense that the string is parsed exactly as if indentation were just not in the source code. Quickly checking multi-line strings in Elixir, it looks like the string is dedented before escapes, and, e.g., `\n` doesn't need any extra indentation space after it to work (Elixir also appends final newline). It may also look weird that `\s` can be used in place of indentation (and mixed with spaces, making the literal string not look evenly indented) in case it is interpreted first. Functions like python's `textwrap.dedent` can remove indentation only after the escaped string has been parsed, and it feels more like a deficiency. As far as the JavaScript proposal goes, IIRC it takes the `String.raw...` form of the string so escapes should not interpreted. In the end, I support escapes after dedenting.

The kdlua implementation expands escapes _before_ dedenting, I think [(link)](https://github.com/danini-the-panini/kdlua/blob/728eb6a928fac3e3746d2ee3ec9452de4fec738d/src/tokenizer.lua#L452).

zkat:
ahhh yes. I see your point. I'll change it back to be dedent-before-escape.

zkat:
There, that's done :)

CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+  * Around `=` for props (`x = 1`)
+* The BOM is now only allowed as the first character in a document. It was
+  previously treated as generic whitespace.
+* Multi-line strings are now automatically dedented, according to the common
+  whitespace matching the whitespace prefix of the closing line. Multiline
+  strings and raw strings now must have a newline immediately following their
+  opening `"`, and a final newline plus whitespace preceding the closing `"`.
+* SMALL EQUALS SIGN (`U+FE66`), FULLWIDTH EQUALS SIGN (`U+FF1D`), and HEAVY
+  EQUALS SIGN (`U+1F7F0`) are now treated the same as `=` and can be used for
+  properties (e.g. `お名前＝☜(ﾟヮﾟ☜)`). They are also no longer valid in bare
```
tjol, line 64: This example doesn't look right: surely `(` and `)` aren't allowed here?

tjol:
Implementing the multi-line string and whitespace escaping rules is proving *quite* subtle.

> When processing a Multi-line String, implementations MUST resolve all whitespace escapes after dedenting the string.

This sounds simple enough: if there are newlines in the string, I check that the indentation is consistent and remove it. Then, I handle the various backslash escapes.

That should take care of illegal strings like
```
  "
  foo \
bar
  baz
  "
```
(from in the spec), and legal strings like
```
"
    Hello
    \
         World
    "
```
(which is equal to `"Hello\nWorld"`, if I'm understanding this correctly)

This algorithm does **not** work for this example in the spec:
```
    "Hello\n\
    World"
```
Before considering the `\` escaping the newline, this looks very much like a syntax error: there is a newline in the string, but there is no initial or final newline. I believe the formal grammar also prohibits this string.

The spec prose *appears* to have a solution to this conundrum: (emphasis mine)
> When a Quoted or Raw String spans multiple lines with literal, **non-escaped** Newlines, it follows a special multi-line syntax ...

So if all newlines in the string are escaped, it is not a multi-line string? To my mind that should imply that escaped newlines are not newlines for the purposes of dedenting, and contradicts the rule that dedenting comes before backslash escapes. Overall, not very satisfying.

I would suggest:
 * removing “, non-escaped” from the definition of multi-line strings, requiring that every string that contains ANY literal newlines follows the multi-line-string rules. This matches what the formal grammar already says.
 * replacing the offending example with 
```
"
Hello\n\
    World
"

**Comment by tjol** [comment]
I wonder if the more intuitive way for strings to work would be:
1. remove escaped whitespace
2. dedent multi-line string
3. resolve other backslash escapes

This should have the same result as the #391 rule for all strings valid under that rule, but also accept more cases with escaped newlines.

**Comment by zkat** [comment]
@tjol sorry for the delay in responding:

I'm confused, what you're describing is definitely the intended behavior. You should still be able to write "multiline" strings by using whitespace escapes, they're just not going to be beholden to the multiline string rules, and can be easily detected by looking for the character sequence `\<NL>`.

But maybe allowing that is indeed too confusing and too painful to implement? So your suggestion means that whitespace escapes essentially no longer work unless you're in multiline string mode?

**Comment by tjol** [comment]
@zkat I think the way the spec is currently written – certainly the way I understood the 'letter of the law' while implementing – whitespace escapes basically can't escape newlines in single-line strings, yeah. But I agree that's probably not the way it *should* work. (Either way should be easy enough to implement)

I'll have another look at the wording and how to maybe clarify it (probably in a few days' time)

**Comment by zkat** [comment]
@tjol  yeah it could definitely be improved. Thanks for being willing to take a look!!

**Comment by tjol** [comment]
Ok - alternative suggestion PR: #392

File: SPEC.md [hunk 37]
```diff
@@ -334,32 +365,236 @@ interpreted as described in the following table:
...
+      foo
+  This is no longer on the left edge
+          bar
+```
+
+Equivalent to `"      foo\n  This is no longer on the left edge\n          bar"`.
+
+-----------
+
+Empty lines can contain any whitespace, or none at all, and will be reflected as empty in the value:
```
tjol, line 524: I don't think this matches the description above or the discussion at https://github.com/kdl-org/kdl/pull/286/files#r1483699764

Should this say something like "Empty lines may omit the whitespace prefix:" instead?

SPEC.md
```diff
@@ -302,29 +273,89 @@ IEEE 754-2008 decimal floating point numbers
...
 
 ### String
 
-Strings in KDL represent textual [Values](#value), or unusual identifiers. A
-String is either a [Quoted String](#quoted-string) or a
-[Raw String](#raw-string). Quoted Strings may include escaped characters, while
-Raw Strings always contain only the literal characters that are present.
+Strings in KDL represent textual UTF-8 [Values](#value). A String is either an
+[Identifier String](#identifier-string) (like `foo`), a [Quoted String](#quoted-string) (like `"foo"`) or
+a [Raw String](#raw-string) (like `#"foo"#`). Identifier Strings let you write short, "single-word" strings with a minimum of syntax; Quoted Strings let you write strings with whitespace (including newlines!) or escapes; Raw Strings let you write strings with whitespace *but without escapes*, allowing you to not worry about the string's content containing anything that might look like an escape.
```
j-tai, line 285: Hey, I like the new barewords! I think it would be useful to document whether the different types of strings MUST be treated identically or not. In other words, we should specify whether the following MUST behave identically:

```
foo arg
foo "arg"
```

If they MAY be treated differently, then this could be useful for CLI-like syntaxes, such as

```
node1 --option  // treated as an "option"
node1 "--option"  // treated as a "positional argument"
```

SPEC.md
```diff
@@ -302,29 +273,89 @@ IEEE 754-2008 decimal floating point numbers
...
+Unicode Scalar Values other than [non-identifier
+characters](#non-identifier-characters).
+
+A handful of patterns are disallowed, to avoid confusion with other values:
+
+* idents that appear to start with a [Number](#number) (like `1.0v2` or
+    `-1em`) or the "almost a number" pattern of a decimal point without a
+    leading digit (like `.1`).
+* idents that are the language keywords (`inf`, `-inf`, `nan`, `true`,
+  `false`, and `null`) without their leading `#`.
```
j-tai, line 309: I wonder if it would be a good idea to make this a bit more conservative? For example, we could also disallow `NaN`, `-nan`, `+inf`, `Infinity`, etc.

SPEC.md
tabatkins: The intention is definitely that they are identical, just different syntaxes for a "string". All the existing kdl1 locations that allow idents and strings (type annotations, node names, property names) don't allow you to distinguish between the two syntaxes, any more than they let you distinguish between the various quoted string syntaxes. Extending this to attribute values and property values shouldn't change things.

If you do have a node where you want to have both string positional args and boolean flags, the recommended way is to make the flags into boolean properties. `node1 --option=#true` and `node1 "--option"` are completely distinguishable. (If you're not mixing these two things, tho, then just having the presence of an ident string indicate a flag is perfectly fine, like `node1 --option`.)

j-tai: Fair enough. In that case, I think it would be worth adding this to the spec. Something like:

"Implementations MUST treat the different string forms identically."

zkat: that's not technically correct. A document-oriented parser might represent those differently, because they're represented differently in text and they want to maintain formatting and allow control of formatting. I'm not sure the spec needs any additional language beyond specifying "these two syntaxes represent the same data type," which is already expressed.

zkat:
This PR has gotten really unwieldy, so I'm going to go ahead and merge it to main. We can continue iterating on v2 off that branch from now on. I've updated the readme to clarify this in case anyone gets confused when first visiting.

Selonian83:
nice work pretty !!!

PR #288, opened 2022-08-30
Use forward slash in solidus-escape test

bgotink:
The forward slash was removed from possible escapes, but the test currently checks that backslash is no longer allowed.


zkat (approved):
Oh haha oops

PR #291, opened 2022-09-01
Add escaped whitespace note to v2 changelog

hkolbeck:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Solidus/Forward slash (`/`) is no longer an escaped character.
 * Single line comments (`//`) can now be immediately followed by a newline.
+* Blocks of whitespace in strings can now be escaped and discarded.
```
zkat, line 9: can you add a bit more detail here?

zkat (changes requested):

CHANGELOG.md
hkolbeck: Certainly, are there any particular details you'd like included?

zkat: I think just expand the sentence a bit to say something like "all literal whitespace after `\` in a string is now escaped and discarded", to be more specific about what the change is?

zkat (approved):

PR #293, opened 2022-09-02
Make JSON-in-KDL more minimalistic

### LINKED ISSUE #281: Better json-in-kdl microsyntax
The current JiK syntax, while functional, feels *very* clunky to me. The original syntax was created before type hints, and while the 2.0.0 specification takes advantage of them, it doesn't *quite* feel right. I'd like to propose a new version of JiK that uses type hints to their full ability, while making JiK easier to both read and write. This isn't a full PR or specification yet because I'd like to hear what other folks think.

TL;DR:
- instead of `-`, `array`, and `object` as node identifiers, use the identifier as the element key
- mark `(array)` and `(object)` using type hints, with no type hint assuming literal
- use `-` for anonymous elements like the outermost object or items in an array

Examples
```json
{
    "foo": 5,
    "bar": [
        true,
        false,
        null
    ],
    "baz": {
        "qux": "thud"
    }
}
```

```kdl
(object)- {
    foo 5
    (array)bar {
        - true
        - false
        - null
    }
    (object)baz {
        qux "thud"
    }
}
```

bgotink:
This PR applies the new JiK microsyntax proposed in #281. Instead of explicitly requiring `(array)` and `(object)` tags as proposed in the OP, the heuristic described in https://github.com/kdl-org/kdl/issues/281#issuecomment-1215058690 is applied to detect the correct type while allowing the tags to be used in two cases where ambiguities arise.

In my comment https://github.com/kdl-org/kdl/issues/281#issuecomment-1215058690 I also proposed ditching the requirement for a single top-level node, but that would limit JiK to encode only JSON arrays and objects while also making it impossible to encode a JSON stream, so I did not apply that here.

Closes #281

### DIFF
--- a/JSON-IN-KDL.md
+++ b/JSON-IN-KDL.md
@@ -3,30 +3,28 @@ JSON-in-KDL (JiK)
 
 This specification describes a canonical way to losslessly encode [JSON](https://json.org) in [KDL](https://kdl.dev). While this isn't a very useful thing to want to do on its own, it's occasionally useful when using a KDL toolchain while speaking with a JSON-consuming or -emitting service.
 
-This is version 2.0.0 of JiK.
+This is version 3.0.0 of JiK.
 
-JSON-in-KDL (JiK from now on) is a kdl microsyntax consisting of three types of nodes:
+JSON-in-KDL (JiK from now on) is a kdl microsyntax consisting of named nodes that represent objects, arrays, or literal values.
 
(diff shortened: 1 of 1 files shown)


LemmaEOF (approved):

tabatkins (approved):

tabatkins:
Very nice, I like what's going on here a lot better.

LemmaEOF:
🎉🎉🎉

PR #298, opened 2022-09-19
Forbid zero-length identifiers.

Patitotective:
## Discussions
https://github.com/kdl-org/kdl/pull/286#issuecomment-1250559170 and https://github.com/kdl-org/kdl/discussions/177.

## Changes
- Removed tests that allowed empty strings as identifiers (`blank_node_type.kdl`, `blank_arg_type.kdl`, `blanl_prop_type.kdl`, `empty_quoted_prop_key.kdl` and `empty_quoted_node_id.kdl`).
- Edited SPEC.md to be clear about not allowing zero-length identifiers.

## Notes
- Need to update grammar as well, not sure how to say _more than 0 string characters_ (create another rule ?).
- Why some tests use `blank` prefix while other `empty`?
- In the SPEC.md, define rename _Identifier_ to _Bare-identifier_ and define _Identifier_ as a _Bare-identifier_ or quoted string (with more than 0 characters).


zkat:
I'm strongly -1 with this change. Identifiers can be arbitrary keys in key/value pairs and this would present a potentially serious interop problem. For example, I'll note that JSON keys can be empty strings as well. This feels arbitrary and conflates the concept of `None`/`null` with "valid but empty value". I don't see a reason to do this.

PR #304, opened 2022-09-28
Remove credit from test suite README

hkolbeck:
I came across this and felt weird having it here, especially when so many other folks have contributed. I'd prefer to be uncredited in this regard.


zkat (approved):

zkat:
If you say so... thanks anyway!!

PR #310, opened 2022-10-10
Replace use of non-defined 'tag' term

### LINKED ISSUE #306: The word ‘tag’ is nowhere defined
The word ‘tag’ is used in a few documents, including the KQL spec and the JSON-in-KDL spec. From context, it appears to mean ‘type annotation’.

The word ‘tag’ is not used or defined in the README, on kdl.dev, or in the KDL spec. What is worse, the website and README do *contain* the word ‘tag’ in this example:

    tag /*foo=true*/ bar=false

suggesting that ‘tag‘ might be a term for the node name.

larsgw:
In the specifications of KQL and JiK, replace the usage of 'tag' with 'type annotation', as that is the term in the main KDL specification.

Close #306

### DIFF
--- a/JSON-IN-KDL.md
+++ b/JSON-IN-KDL.md
@@ -41,14 +41,14 @@ Arguments and children can be mixed, if desired. The preceding example could als
 }
 ```
 
-Two otherwise-ambiguous cases must be manually annotated with an `(array)` tag:
+Two otherwise-ambiguous cases must be manually annotated with an `(array)` type annotation:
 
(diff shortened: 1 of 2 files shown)


zkat (approved):

larsgw:
Would you consider labeling this PR `hacktoberfest-accepted`? I really like their t-shirts, in spite of the problems Hacktoberfest has (and in spite of friends & family generally being confused by the text.)

PR #312, opened 2022-10-17
Add Intellij IDEA plugin to README.md

Exidex:


README.md
```diff
@@ -54,6 +54,7 @@ entirety, but in the future, may be required to in order to be included here.
 
 * [VS Code](https://marketplace.visualstudio.com/items?itemName=kdl-org.kdl&ssr=false#review-details)
 * [vim](https://github.com/imsnif/kdl.vim)
+* [Intellij IDEA](https://plugins.jetbrains.com/plugin/20136-kdl-document-language)
```
djmattyg007, line 57: This looks like an alphabetically sorted list, which means the IntelliJ plugin should be first, not last.

zkat: it's not really

zkat (approved):

PR #334, opened 2023-09-25
Add link to Go implementation

lunjon:
I've attempted to write a parser implementation in Go. It is not entirely complete, but almost.

Would it be okay to add as implementation?

Thanks in advance!


zkat:
Yes! Please also add it to the site over at https://github.com/kdl-org/kdl-org.github.io

lunjon:
> Yes! Please also add it to the site over at https://github.com/kdl-org/kdl-org.github.io

Thank you!

Absolutely, will do.

PR #336, opened 2023-10-16
Add kdl-go to README.md

sblinch:
I've just released a complete KDL library for Go -- hoping to have it added here. It passes all test cases and includes marshal/unmarshal support similar to equivalent Go JSON/YAML/TOML libs.


zkat (approved):
Awesome! Please be sure to make the same PR over at https://github.com/kdl-org/kdl-org.github.io so it shows up on the site, too. :)

PR #382, opened 2024-02-18
Ensure spec allows slashdash right after node separator

bgotink:
As mentioned in https://github.com/kdl-org/kdl/pull/286#discussion_r1493297090


zkat (approved):
ah nice. This should work!

PR #383, opened 2024-02-19
Update README.md

wommy:
swap neovim integration link 
from tree-sitter project to the specific language grammar


zkat:
Could you simply add this one, instead of removing the other, so folks know about both? Is there a strong reason for removing the current one entirely?

spaarmann:
I would also support at the very least keeping the generic nvim-treesitter link as that is what you have to actually install as a user to get editor support for kdl (plus the usual nvim-treesitter config to enable a language).

zkat (approved):

PR #388, opened 2024-04-14
kql: only allow top() at start of selector

alightgoesout:
Following discussion #387, I propose a change to the KQL grammar. With this change, `top()` will only be allowed at the start of a KQL selector.

### DIFF
--- a/QUERY-SPEC.md
+++ b/QUERY-SPEC.md
@@ -30,6 +30,11 @@ properties, node names, etc). With the exception of `top()` and `()`, they are a
 used inside a `[]` selector. Some matchers are unary, but most of them involve
 binary operators.
 
+The `top()` matcher can only be used as the first matcher of a selector. This means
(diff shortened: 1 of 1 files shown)


zkat:
Maybe we can add some explanatory text as well, to clarify that `top()` is toplevel only, not just put it in the grammar?

alightgoesout:
Yes you're right, I have added an explanation.

zkat (approved):

zkat:
lgtm, I'll let @larsgw have an eyeball at it since he was part of our discussion, too.

larsgw:
I'd rather have `selector` and `selector-subsequent` so that the main selector rule is called `selector` and not `selector-start`, that makes it slightly easier to translate to a grammar. Otherwise, lgtm.

alightgoesout:
I have renamed the rules as suggested.

larsgw (approved):

PR #392, opened 2024-06-13
[v2] Tweak rules for escaped whitespace in multi-line strings

tjol:
These rules are a bit more liberal than what was described previously, but I think they're clearer and more consistent:

 * This way, strings have the (I think intuitive) property that, when you 'blindly' remove the whitespace escapes, the meaning is unchanged.
 * If you take any valid single-line string and add a newline character and some indentation both at the start and the end, the string will still be valid (and unchanged) - previously, this was not necessarily the case if there were whitespace escapes.


zkat (approved):
I like this! Thank you!

Issue #399, opened 2024-10-05
[v2] Remove = normalization


zkat:
See also: #386 

Since we decided that fancy quotes were "too much", I think we should get rid of = normalization for the same reason. It's not a huge deal if we keep it, but I think it's best to keep this kind of thing simple and unsurprising, in the end.

PR #400, opened 2024-10-05
get rid of syntactically significant unicode equals signs

zkat:
Fixes: #399


SPEC.md
```diff
@@ -133,14 +133,11 @@ make it act as plain whitespace, even if it spreads across multiple lines.
...
 
-Any of the following characters may be used as equals signs in properties:
+Only the following character may be used as equals signs in properties:
 
 | Name | Character | Code Point |
 |----|-----|----|
 | EQUALS SIGN | `=` | `U+003D` |
-| SMALL EQUALS SIGN | `﹦` | `U+FE66` |
-| FULLWIDTH EQUALS SIGN | `＝` | `U+FF1D` |
-| HEAVY EQUALS SIGN | `🟰` | `U+1F7F0` |
```
larsgw, line 143: Is it necessary to keep this section? It's has only one reference, in the previous section, that could be replaced with "equals sign (`=`, code point `U+003D`)". Regardless, the previous section might need to be reworded anyway (#401).

zkat: done

larsgw (changes requested):
Just realized that the example in the readme is now outdated:

> ```
> // And you can also use unicode, even for the equals sign!
> ノード　お名前＝☜(ﾟヮﾟ☜)
> ```

Issue #401, opened 2024-10-06
optional-node-space around equals sign in properties


larsgw:
The grammar for a property looks like this at the moment:

```
prop := string optional-node-space '=' optional-node-space value
```

However, the text says:

> A Property is composed of a [String](https://github.com/kdl-org/kdl/blob/0c04750006c353786555e980ef8fcd77e99c2904/SPEC.md#string), followed immediately by an [equals sign](https://github.com/kdl-org/kdl/blob/0c04750006c353786555e980ef8fcd77e99c2904/SPEC.md#equals-sign), and then a [Value](https://github.com/kdl-org/kdl/blob/0c04750006c353786555e980ef8fcd77e99c2904/SPEC.md#value).

Probably the latter needs to be reworded to allow whitespace, esclines, and slashdash comments.

(Sidenote, I'm a bit unsure if slashdash comments are good in this context. Firstly, `node prop= /-"foo" "bar"` would be equivalent to `node prop="bar"` which is not very obvious (it looks like `"foo"` and especially `"bar"` are just Values to me on first sight). Secondly, `node prop=/-"foo" "bar"` is invalid which seems a bit inconsistent, I feel like that would be clearer. ~Thirdly, I think `node prop /-="foo" ="bar"` is probably very annoying to parse, especially now `prop` is a valid Value.~)

larsgw:
Also, I think `node prop /-foo = bar` is ambiguous, either a Node with a Value `"prop"` or with a Property `prop=bar`. Given that, we might need to change the grammar as well.

zkat:
I think a good test for where `/-` should go is whether we would also allow comments in there.

Would `prop = /-foo bar` not make sense, but `prop = /*foo*/ bar would? Why? I feel like the latter is very sensible (and imo, so is the former)

zkat:
The latter is definitely a bigger concern. I'm not sure what to do about that ambiguity: you definitely should be able to slashdash entire props...

larsgw:
Maybe a better test for where `/-` should go, is whether the entity you're commenting out could occur there without being commented out. i.e. you already cannot put a slash-dashed Node between two Values, or between a Value and a Children block. So I would suggest no slash-dash before a `=` in a prop, and after a `=` only allow slash-dashed Values, not Properties. I would also prefer to remove the mandated whitespace between `=` and `/-` in the current grammar.

Does this work?

```
prop-value-space := plain-node-space* ('/-' plain-node-space* value plain-node-space+)*
[...]

PR #407, opened 2024-11-27
[v2] more predictable slashdash

zkat:
Fixes: https://github.com/kdl-org/kdl/issues/401


zkat:
also worth noting: This restricts `/- { child1 }` to only being present _after_ props and arguments, so the following is not legal:

```kdl
node /-{child1} foo;
```

but this is:

```kdl
node foo /-{one} {two} /-{three}
```

I believe this is what you were hoping for, @tabatkins?

SPEC.md
```diff
@@ -764,25 +765,29 @@ document := bom? nodes
...
-optional-node-space := node-space*
+// Whitespace within nodes, where newline-ish things must be esclined.
+node-space := ws* escline ws* | ws+
 
-base-node := type? optional-node-space string (required-node-space node-prop-or-arg)* (required-node-space node-children)?
-node := base-node optional-node-space node-terminator
-final-node := base-node optional-node-space node-terminator?
-node-prop-or-arg := prop | value
+base-node := slashdash? type? node-space* string
+      (node-space+ node-prop-or-arg)*
```
tabatkins, line 777: Just for consistency, I'd move the optional `slashdash` to here, rather than putting it into `node-prop-or-arg`.

tabatkins (approved):
Very minor style nit, otherwise r+

SPEC.md
zkat: Sounds good. Additionally, I just realized I didn’t put any space allowances around the /-. I’m thinking of making it like:

`slashdash := '/-' line-space*`

Which means that these would both be legal:

```kdl
// Very reasonable, IMO
/-
node 1 2 3

// interesting and I think acceptable—it means you can use them like esclines:

node 1 /-
2 3
```

We can totally ban the second behavior but I think it’s better than having to special case allowing esclines after the /-. This way, we just say “/- consumes all following line-space, and the next “item”, regardless of where you put it (as long as it’s a legal /- location“. 

Interestingly, that implies the following are legal too:

```kdl
node /- //next thing is commented
  foo

node /- /* commented */ foo
```

I’ll have to adjust the spec prose too, to clarify that these cases are possible.

PR #410, opened 2024-12-01
[tests] [bug] test unicode_silly's input data is a .kd file

Alch-Emi:
Hi!  I think the file extension for this test is wrong.

The input file for the `unicode_silly` test has the file extension `.kd` instead of `.kdl`.  The output file seems fine though!


tabatkins:
Thanks for the fix!

PR #412, opened 2024-12-02
[bug] [tests] The expected output of unicode_silly contains a non-normalized space

Alch-Emi:
Follow-up to #410, I think that the ideographic space between the node and the property in `unicode_silly` should be normalized to a normal space in the expected output, but please let me know if I'm misreading!


zkat (approved):

zkat:
yep! Nice catch :)

PR #417, opened 2024-12-06
Make multi-line a top-level string type; demote Raw String.

### LINKED ISSUE #416: String spec sections are little confusing now due to multiline
The string section of the spec <https://github.com/kdl-org/kdl/blob/main/SPEC.md#quoted-string> divides into three subsections: ident strings, quoted strings, and raw strings. Multiline strings are a fourth subsection after that, because both quoted and raw can have multiline.

The quoted string section still talks about allowing literal newlines, and says that they "follow specific Multi-line String rules", but this is no longer correct. Multilines have a different syntax, and single-lines no longer allow (unescaped) literal newlines.

At the same time, "raw" strings feel much more like a property that can be applied to strings, rather than a type of string on their own.

I think it would flow a lot better to instead separate strings into ident/quoted/multiline, and then define raw-ness as a property that can apply to both quoted and multiline.

(I'll put up a PR in a bit.)

tabatkins:
Also lightly rewrite some bits to make more sense with this new organization. 
Fixes #416

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -276,24 +276,27 @@ node prop=(regex).*
 ### String
 
 Strings in KDL represent textual UTF-8 [Values](#value). A String is either an
-[Identifier String](#identifier-string) (like `foo`), a [Quoted
-String](#quoted-string) (like `"foo"`) or a [Raw String](#raw-string) (like
-`#"foo"#`):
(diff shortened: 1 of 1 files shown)


tabatkins:
Merging since there were :+1:s in the issue; feel free to tweak.

PR #418, opened 2024-12-06
Indicate expected-fail tests with a _fail suffix.  Fixes #365

### LINKED ISSUE #365: Indicate invalid tests in filename?
Currently, if you're reviewing the test suite and want to know if a given test is meant to pass or fail, you have to go check the `expected/` folder to see if there's an output file there or not. That's slightly annoying. Could we *also* indicate that something is meant to fail in the filename? Like a `_FAIL` at the end or something.

Then we can add a lint that verifies that every test ending in `_FAIL` lacks an expected file, and vice versa, to make sure these two signals don't drift out of date.

tabatkins:
* Renamed the existing expected-fail tests to have a _fail suffix, as suggested in #365.
* Edit the README in /tests to document the requirement to use that suffix.
* Added a lint to pushes and PRs that touch the tests/ folder, to ensure that the suffix is used (and, while I'm there, that there are no "orphaned" expectation files).


zkat:
bless u for doing this🙏🏻🙏🏻🙏🏻

PR #419, opened 2024-12-07
Allow quotes in multi-line strings

### LINKED ISSUE #415: [v2] Should triple-quoted strings be allowed to contain unescaped quotes?
I rather like the new multi-line string syntax (#413). It's elegant, it's familiar, it fixes a thing we were worried about, but — it raises a question:

```
"""
is "this"
a valid string?
"""
```

The draft spec says no right now, but there's really no reason not to allow `"` characters in a triple-`"` string (as long as there aren't three in a row). Plus, Python allows it.

tjol:
Fixes #415.

This changes the grammar and prose to allow (multi-line) strings to include unescaped quotes as long as they don't contain their own closing delimiter.

I haven't touched the grammar for raw strings since it already allows raw strings to contain `"` and `#` in any order and number. The restriction that the body of a raw string can't contain the corresponding closing delimiter was never explicit in the grammar (afaict), and I don't think it's worth making explicit there when it's already explicit in the prose.

Also: added tests, tweaked the wording for raw string.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -430,7 +430,9 @@ before the closing `"""`.
 All in-between lines that contain non-newline characters
 _MUST_ start with _at least_ the exact same whitespace as the final line
 (precisely matching codepoints, not merely counting characters or "size");
-they may contain additional whitesapce following this prefix.
+they may contain additional whitesapce following this prefix. The lines in
+between may contain unescaped `"` (but no unescaped `"""` as this would close
+the string).
 
 The value of the Multi-Line String omits the first and last Newline, the
 Whitespace of the last line, and the matching Whitespace prefix on all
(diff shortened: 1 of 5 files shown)


SPEC.md
```diff
@@ -653,7 +654,7 @@ You can show examples of """
 without worrying about escapes.
 ```
 
-or equivalently, `#"You can show examples of """\n    multi-line strings\n    """\nwithout worrying about escapes."#` as a Quoted String.
+or equivalently, `"You can show examples of """\n    multi-line strings\n    """\nwithout worrying about escapes."` as a Quoted String.
```
zkat, line 657: The stuff between backticks is how you would represent the contents as a single line string. I think you’ll want to escape the quotes here.

SPEC.md
```diff
@@ -856,7 +857,7 @@ disallowed-keyword-identifiers := 'true' - 'false' - 'null' - 'inf' - '-inf' - '
 
 quoted-string := '"' single-line-string-body '"' | '"""' newline multi-line-string-body newline unicode-space*) '"""'
 single-line-string-body := (string-character - newline)*
-multi-line-string-body := string-character*
+multi-line-string-body := (('"' | '""')? string-character)*
```
zkat, line 860: [chefkiss] I love how clean a change this is

zkat (approved):
This looks good! I think there’s a minor thing to look at but I also don’t mind merging as is.

zkat:
Thanks a bunch for taking the time to do this!

SPEC.md
tjol: Fixed

Issue #423, opened 2024-12-08
Rules about raw strings?


IceDragon200:
## And then he tried updating his code...

So while getting kuddle all caught up with the spec changes (draft.6 as of this writing), I ran into this lovely test case (outside of the challenge that is slashdash right now):

> [multiline_raw_string_single_line_err_fail.kdl](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/input/multiline_raw_string_single_line_err_fail.kdl)
```kdl
node #"""one line"""#
```

At a glance, yeah it should fail, it's intend to be a multiline string... right?

Well yes, but no, raw strings can contain quotes as evidenced by this fun little parser bendy test (thanks to whoever made this gem, I hate it lol):
> [raw_string_just_quote.kdl ](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/input/raw_string_just_quote.kdl)
```kdl
node #"""#
```

So then, why does the former case count as a failure?
Or rather why am I even bringing this up?

## The purgatory known as multiline (raw) strings

In the former case, the multiline condition requires that the triple quote be followed by its newline (\n and others) or newline sequence (\r\n and friends).
In this case, it does not, therefore it should rightfully be treated as a single line raw string, so in that case:
```kdl
node #"""one line"""#
// Is now equivalent to:
node "\"\"one line\"\""
```

This is important to note because it's the same reasoning that makes the latter test work in the first place:
```
// for example, kuddle would fail on the below line since it had a strict adherence to multiline strings being followed by newlines (any triple quotes would trigger a multiline check, regardless of whether or not the closing quotes were present, since the string is scanned character by character)
node #"""#
// Is equivalent to:
node "\""
```

## Regular quoted strings do not suffer from this

Because you can only have quotes in a multiline string within that context (it makes sense there since you need to terminate on """ and leaving out single " would feel weird).

## Did this really need to be an issue, why not just PR and fix the issue?

In hindsight possibly no, but I wanted to raise an issue so others could see it and maybe we discuss the fun topic of raw strings.

zkat:
Huh. I would expect `#"""#` because the rule is very much “any opening tells triple quotes MUST be followed by a newline”. It may just be a test that needs updating tbh

zkat:
Also, by that rule, `#"""foo"""#` is a syntax error too

tjol:
[...]

PR #425, opened 2024-12-09
Clarify raw single-line string grammar

tjol:
See #423 

I hate this but I think it might be correct


zkat (approved):
thanks, I hate it

I'm gonna leave this just for a little to see if anyone else can eyeball the grammar and make sure it works (it looks good to me?)

This is what we wanted, though. Thank you :P

tabatkins:
I think this is right, and not all that nasty actually. It's basically the same "no masquerading as a number" construction we have for ident strings.

Issue #426, opened 2024-12-11
Slashdash, per grammar, doesn't allow an escaped newline after it


tabatkins:
Per the current grammar, the test `slashdash_arg_before_newline_esc.kdl` contains a parse error.

```kdl
node /-    \
    arg
```

Notably, slashdash allows `line-space` at its end, which does *not* include `escline`. The escaped newlines is allowed *before* the slashdash or *after* the attribute (since then it would be part of the allowed `node-space`), but not between the slashdash and the attribute.

Given that this test is intended-successful, I suspect it's intended that a slashdash allow `node-space` after itself too. Does this sound correct?

zkat:
oh. huh.

I was curious as to why `kdl-rs` wasn't failing this test, and it's because... yeah, [I'm effectively putting a `node-space*` before every `node-entry`](https://github.com/kdl-org/kdl-rs/blob/main/src/v2_parser.rs#L418-L440)

So yeah, `slashdash := '/-' (node-space | line-space)*` seems like a good patch. I'm surprised no one else has run into it either, but I assume that's a function of how new this change is, and how... realistically, most parsers would also allow `node-space*` before node entries.

Alch-Emi:
Maybe this isn't intended behavior, but shouldn't this be allowed under the current (75a0869) grammar?

The entries part of the spec is:

```ebnf
node := ... (node-space+ slashdash? node-prop-or-arg)* ...
```

and `node-prop-or-arg` is

```ebnf
node-prop-or-arg := prop | value
value := type? node-space* (string | number | keyword)
```

So if `type?` is `""` and `node-prop-or-arg` is `value`, then the collapsed expression should look like:

```ebnf
node := ... (node-space+ slashdash? node-space* (string | number | keyword))* ...
```

Which does permit `node-space` after the `slashdash`, even if it's not the `linespace` that's directly associated with the `slashdash`.

zkat:
@Alch-Emi ahh good catch. Yeah. Having it only work when you don't have a type annotation is supremely weird lol

anyway this is pretty minor. I'll PR it and probably self-merge.

tabatkins:
@Alch-Emi oh lol nice catch, that means I'm doing something wrong still

PR #427, opened 2024-12-11
Allow node-space after slashdash and add a bunch more slashdash tests

zkat:
Fixes: https://github.com/kdl-org/kdl/issues/426

I went ahead and added a few more tests that I think would be reasonably expected to fail. We'll see what happens when I try to actually plug them into kdl-rs :P


tests/test_cases/input/slashdash_escline_before_node.kdl
```diff
@@ -0,0 +1,3 @@
+/-\
```
tjol, line 1: Based on the way ckdl failed this (now fixed), is it possible this is the only test for a line continuation outside of a node?

zkat: oh it might be. Although I _feel_ like I saw at least one that had leading esclines before the first node. Would be nice to have plain escline tests with nothing else.

PR #430, opened 2024-12-12
Make the raw-string productions non-greedy, and describe the infallibility

### LINKED ISSUE #415: [v2] Should triple-quoted strings be allowed to contain unescaped quotes?

tabatkins:
Closes #415

Doing this as a PR to get another eye on it, otherwise I'll merge it myself in a bit.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -872,10 +872,10 @@ escape := ["\\bfnrts] | 'u{' hex-digit{1, 6} '}' | (unicode-space | newline)+
 hex-digit := [0-9a-fA-F]
 
 raw-string := '#' raw-string-quotes '#' | '#' raw-string '#'
-raw-string-quotes := '"' single-line-raw-string-body '"' | '"""' newline multi-line-raw-string-body newline unicode-space* '"""'
-single-line-raw-string-body := '' | (single-line-raw-string-char - '"') single-line-raw-string-char* | '"' (single-line-raw-string-char - '"') single-line-raw-string-char*
+raw-string-quotes := '"' single-line-raw-string-body '"' | '"""' newline multi-line-raw-string-body '"""'
+single-line-raw-string-body := '' | (single-line-raw-string-char - '"') single-line-raw-string-char*? | '"' (single-line-raw-string-char - '"') single-line-raw-string-char*?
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -928,6 +928,11 @@ Specifically:
   (`\\`).
 * `*` is used for "zero or more", `+` is used for "one or more", and `?` is
   used for "zero or one".
+* `*?` (used only in raw strings) indicates a *non-greedy* match.
+    It also indicates *infallibility*, with a scope of the `string` production:
+    once it successfully matches enough characters to satisfy that production
+    the first time, it is not allowed to backtrack and continue matching further,
+    even if that results in a parse failure.
```
Alch-Emi, line 935: Is this concept of infallibility a standard thing?  I'm a little new to the world of practical parsers, but I think this is my first time hearing it.  Having read the discussion in #​415 I think I understand what you mean (once the closing delimiter is parsed, that *must* be the closing delimiter for the string), but I think if I were going off of reading the grammar key alone, I wouldn't be able to understand.

In particular, I think the notion of success would confuse me.  Could parsing a single character count as success?  Has it succeeded once it's parsed until a single double-quote?  At which point is the parse locked-in?

If this is a standard concept that I just haven't been exposed to yet, I think it could be really helpful to link to an external document that explains it in more detail, maybe even with examples.  And if it's not, do you think you could go into a little more detail in this section so that folks who are new to the parsing scene (like me!) might have an easier time with it?

zkat: Yeah, “unrecoverable” errors are a common concept in parsing. Sometimes you know there’s absolutely nothing else something could be: https://docs.rs/winnow/latest/winnow/combinator/fn.cut_err.html

tabatkins: Yeah I think I should use the term "cut", that is more common in the literature. Let me reword.

zkat (approved):

PR #431, opened 2024-12-12
went over tests and tweaked a few things

zkat:


tests/test_cases/input/eof_after_escape.kdl
```diff
@@ -1 +1 @@
-node \
+node \
```
tabatkins, line 1: Missed this case, NICE

zkat: oh? You mean this failed in kdl-py? haha nice

tjol: Yeah this also failed in ckdl :-)

PR #432, opened 2024-12-12
More tests for whitespace in multi-line strings

tjol:
@tabatkins please check that I got these right.


tests/test_cases/expected_kdl/slashdash_newline_before_node.kdl
```diff
@@ -1 +0,0 @@
-
```
tabatkins, line 1: Looks like you accidentally included an old version of this expectation; current testsuite has a newline in it.

tjol: Ah, I somehow thought the file had two newlines (and removed the second one).

tjol: Oooh, the file contains CR LF, not just LF. (That's why my test is failing)

That can't be on purpose, can it?

tabatkins (approved):

Issue #429, opened 2024-12-12
`multiline_string_whitespace_only` test wrong?


tabatkins:
Per spec:

> Empty lines can contain any whitespace, or none at all, and will be reflected as empty in the value:

The example following that line just shows off a completely empty line (lacking the required ws prefix), but [`multiline_string_whitespace_only`](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/input/multiline_string_whitespace_only.kdl) tests a string (the last one) that *does* contain the required prefix, and then some extra whitespace after. The [expectation file](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/expected_kdl/multiline_string_whitespace_only.kdl) expects the additional whitespace to be reflected into the string's value.

I think it shouldn't be doing that - the expectation should have the final string just contain a newline and nothing else.

If the current test expectation is correct, then the spec is ambiguous - what happens when a line just contains *some* of the final-line whitespace, but not all? What if it starts with a different whitespace entirely?

(This test was recently added by @tjol; I expect it was just an error.)

zkat:
Yeah that looks to me like it should have just the single newline in it?

zkat:
Also: I think the spec is pretty clear. You can have literally any whitespace in empty lines, and they will always collapse to a single `\n` (unless it’s the penultimate line, in which case it collapses to nothing)

tabatkins:
Agreed that the spec is clear, I just like having another eye on test changes unless it's *incredibly* obvious it's an error. ^_^

And with that, kdlpy's main branch now passes 100% of the v2 tests, woo!

tjol:
I must have missed that detail.

There's a bit of a tension between

> All in-between lines that contain non-newline characters _MUST_ start with _at least_ the exact same whitespace as the final line

and

> Empty lines can contain any whitespace, or none at all, and will be reflected as empty in the value

in the spec. Maybe "non-newline" should be replaced by "non-whitespace"?

So is the idea that lines can have trailing whitespace if and only if they contain non-whitespace characters? I find that mildly surprising. (only mildly)

I think I'll add a few more tests for different permutations of whitespace and non-whitespace.

zkat:
@tjol yeah, maybe just change that to:

> All in-between lines that contain non-newline, non-whitespace characters MUST start with at least the exact same whitespace as the final line

tabatkins:
[...]

PR #434, opened 2024-12-15
Release KDL 2.0.0

zkat:
This is it, folks! I think we're ready to finalize 2.0, after something like 3 years of work.

I'm very happy with the language we've defined here, and how nice it feels to use. I want to express my thanks to everyone who has contributed ideas and discussion and work to its specification, all the various implementers who have helped validate it, and all the users of 1.0 who helped give real, experience-based feedback on the previous version of the language, helping us figure out what could be improved.

I'm gonna give folks a few days to review things and make a decision, then I'm hoping to have consensus from at least a few of the major contributors and implementers before publishing this.

Thanks again, everyone. This wouldn't have been possible without you, and I think we've ended up with a real, shining gem.

NOTE: If you're "just" a community member, your opinion is still welcome, although contributors/implementers will likely be prioritized at this point.


zkat:
heh. The draft.8 tests killed kdl-rs multiline parsing and I realized the grammar wasn't actually allowing escapes on the closing multiline quoted string line.

so there's tests for that too now.

zkat:
@tabatkins @tjol I regret to inform you that the Really Complicated whitespace-only multiline string test was actually wrong (I'm pretty sure).

The prefixes didn't match, and prefixes need to match exactly _before_ any normalization or empty line collapse, unless the line is _completely_ empty.

That is, this should fail, I believe, wince line 2 isn't _completely_ empty, but its prefix doesn't match line 3:

```kdl
"""\n
\t\s\n
\s\s"""
```

tjol:
@zkat I thought we'd established that the rule that empty lines can contain “any” whitespace takes priority over the prefix requirement.

https://github.com/kdl-org/kdl/issues/429#issuecomment-2538160113

zkat:
Ohhh. Uggghhhhhhhhhhhh ok I'll roll this back

eilvelia:
I think the interactions of multiline strings with escapes are pretty weird currently.

> When processing a Multi-line String, implementations MUST dedent the string _after_ resolving all whitespace escapes, but _before_ resolving other backslash escapes.

`\"` should also be interpretered before dedenting (I assume `\"""` doesn't close the string) since otherwise it wouldn't be possible to find the correct enclosing `"""` and deduce the prefix to dedent; if you resolve any escapes, you want to interpret `\\` as well. However, this cannot just be done in two separate phases, or `\\t` will be transformed as `\\t` -> `\t` -> `<tab>`. I think instead it should be noted that, to find the whitespace prefix, at least `\"` and `\\` should be analyzed, and then escapes are resolved from the raw form (i.e. during dedenting the escapes are lexically analyzed but not resolved/saved). It would be nice to add a test with `\"""` inside a multiline string (I can't find one).

> while the following example is allowed
> ```
>   """
>   foo \
> bar
>   baz
>   \   """
> ```

IMO having non-literal whitespace in the prefix-defining line also feels pretty weird. Dedented strings are mostly for readability in the source code, transforming the indentation goes against that. Swift's [multiline strings](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/) are quite similar, and I think can be an inspiration for kdl's strings. Particularly, the final line (or any ws prefix) there can consist of literal whitespace only, the newline escape is not allowed as the last `\n`, lines must either be completely empty or include the prefix.

The "_any_ whitespace in non-content lines" rule forms an exception to the prefix being identical (or at least missing) in all lines, which is also isn't good I think and more difficult to implement (checking the whole line for any whitespace/content instead of only comparing the first characters to the intended prefix). I initially missed this rule when I was updating the ocaml implementation.

zkat:
@eilvelia `\\t` would transform to `t`. Regular escapes are processed before any dedenting happens, and that includes white space escapes. `\"` can never be an issue because it’s always interpreted as simply `"`. `\` doesn’t do its “slurping” behavior unless it’s immediately followed by whitespace.

As far as the Swift rules you mention: it does seem like those would simplify the rules for multi line strings. They would be more limiting but these corner cases would be way less confusing to think about by just… banning them altogether (and simplifying parsing)

I’m curious what others think.

eilvelia:
> @eilvelia \\t would transform to t. Regular escapes are processed before any dedenting happens, and that includes white space escapes. \" can never be an issue because it’s always interpreted as simply ". \ doesn’t do its “slurping” behavior unless it’s immediately followed by whitespace

I think you misunderstood me. `\\t` would be an issue if one naively transforms `\\` to \ and \t to tab in separate phases. You can't transform only the whitespace escape since then `foo\\  bar` would (unexpectedly) activate the whitespace escape. The issue with  `\"` is that this should be allowed:
```
"""
\"""
"""
```
(That would be the most intuitive, also what Swift does and what the grammar and spec currently suggest, as I think.)
If you conpletely disregard all escapes other than the whitespace one before dedenting (as the quote suggests), the second `"""` would be parsed as the string end (and then fail with non-ws final line).

eilvelia:
> Regular escapes are processed before any dedenting happens

Well, per this line, the escapes (other than ws) are resolved _after_ dedenting:

> When processing a Multi-line String, implementations MUST dedent the string after resolving all whitespace escapes, but before resolving other backslash escapes.

zkat:
@eilvelia what that is intended to mean is that:

```kdl
"""
\s\s\s\sfoo
    """
```

is an invalid string (that is, `\s` isn't resolved by dedenting time, so it doesn't actually "count" as whitespace)

Additionally:

```kdl
"""
\"""
"""
```

Is certainly allowed, and certainly is passing _my_ tests right now. The closing `"""` parsing is done before any of the dedenting stuff even applies. That's just standard delimited string stuff.

eilvelia:
Yes, well, that's what should happen, I think the spec is somewhat vague there. That line in particular seemingly suggests that all escapes except ws are unprocessed in case dedent has not been done yet. IIRC the spec only says how escapes are resolved, not how they are lexically analyzed.

zkat:
@eilvelia do you have any thoughts on what kind of rewording would be helpful here?

SPEC.md
```diff
@@ -576,13 +577,14 @@ multi-line """[\n]
...
-Furthermore, a whitespace escape that attempts to escape the final line's newline
-and/or whitespace prefix is invalid since the multi-line string has to still be
-valid with the escaped whitespace removed.
+Multi-line strings support the same mechanism for escaping whitespace as Quoted
+Strings.
+
+When processing a Multi-line String, implementations MUST dedent the string
+_after_ resolving all whitespace escapes, but _before_ resolving other backslash
+escapes. Furthermore, a whitespace escape that attempts to escape the final
+line's newline and/or whitespace prefix is invalid since the multi-line string
```
bgotink, line 586: I'm confused by this phrasing. It seems to say that escaped newline and/or whitespace are not allowed as the multi-line string has to be valid after the escaped whitespace is removed, but the text below and two of the new tests actually allow escaped whitespace in the final line.  
This seems to imply that an escaped newline and/or whitespace are only not allowed if the multiline string would become invalid, but that's not how I read this sentence.

Note that would also allow escaped newlines in some cases, e.g.

```kdl
node """
  lorem
  ipsum
  \
    """
```

which would be equivalent to

```kdl
node """
  lorem
  ipsum
  """
```

which would imo be worth adding as test.

tjol:
@eilvelia Good point about `a\\ b` not resolving to `a\b`. I think it should be clear that the `\ ` is not an escape it this case because the `\` itself is escaped; maybe this could be clearer in the prose. Either way, I'm adding a test for this.

SPEC.md
zkat: @bgotink [this test sort of tests that](https://github.com/kdl-org/kdl/blob/main/tests/test_cases/input/multiline_string_whitespace_only.kdl#L5) but having a more explicit one would be good.

I believe the bit that says that a trailing `\` is invalid actually _intended_ to refer to a case like this:

```kdl
"""
lorem
ipsum\
"""
```

which would be invalid because:

```kdl
"""
lorem
ipsum"""
```

This should probably be reworded, assuming these are the semantics we want to keep.

I'm not terribly inclined to change multiline strings any further, though, tbh.

bgotink: > I'm not terribly inclined to change multiline strings any further,

Oh no, I'm definitely not asking to change multiline strings! I was just confused about the text vs tests.

zkat: oh no that's fine. I was also mostly responding to the ongoing convo with @eilvelia who brought up a lot of good points, though I'm leading towards "clarify, don't change"

IceDragon200:
So I was trying the new tests specially the multiline one: `multiline_string_whitespace_only.kdl`

The test is invalid as last raw string has the incorrect multiline spacing at the final line:

```kdl
// This file deliberately contains unusual whitespace
// The first two strings are empty
node """
    """ """
   \
             
   """ """
                            
 """\
    \ // The next two strings contains only whitespace
    """
       
       
      \s 
    """ #"""
    

  """#
```
  
Looks fine right?
```
  #"""
    

  """#
  
// Is actually
#"""
\u00A0\u205f\u2009\u2000

\u0020\u0020"""#
```

The final line is a mismatch causing the entire test to fail.
Yes the second line is just empty (no characters outside of the newline), which tests the empty line behaviour.

zkat:
I'm pretty happy with where this is at right now. I still need to update kdl-rs with these last couple of changes to make sure everything looks good, but barring any other issues getting brought up, I intend to merge this PR and tag the official 2.0.0 sometime on Saturday.

/cc @larsgw and @tabatkins who I don't think have responded here yet, wanna make sure y'all see this.

larsgw:
I felt like I could not give meaningful feedback without attempting to implement it, but I have unfortunately been very busy with my studies the past few weeks. I don't know if I will before Saturday, but that's fine with me.

zkat:
hmmmmmmmmmm

I was thinking about `#"""#` and how we made it an error now and... I feel weird about it. It's literally the only string that can't be represented by raw strings, at all.

The only way to represent this is to do either:

```kdl
"\""
//or
#"""
"
"""#
```

Which feels really strange? I know it's a relatively minor thing, but it leaves a bad taste in my mouth for something to be, actually, _unrepresentable_ in one of our string syntaxes.

tjol:
That's not the only string you can't represent with raw strings. Another example is `"\r\n"` (you can't do this in a single-line raw string because of the newline and you can't do this in a multi-line raw string because of newline normalization.) You also can't represent `"\u{feff}"` as a raw string (or other strings containing disallowed literal code points). Granted, compared to `"\""`, the other examples I can think of are a lot ... weirder.

zkat:
oh that's a good point. And it does make me feel better. 🤔🤔🤔

eilvelia:
Isn't it a little weird that newlines are allowed after `/-` (the only `line-space` _inside_ a node)?

zkat:
That’s intentional!

zkat:
That is: I thought it would be good for /- to mean “slurp up any and all whitespace until the item being commented out”, as opposed to giving it special rules within nodes. It’s more for simplicity (a single /- definition) than thinking this is a thing that’ll be done all the time. We used to have much more strict locations for /- and it turned out to actually complicate grammars and parsers more than it was worth

eilvelia:
Well, currently slashdash is a node/children/arg/prop "modifier" (as I described in https://github.com/kdl-org/kdl/issues/401#issuecomment-2400657415) and can only be inside a node (including at the beginning of it), changing its behaviour would be as simple as

```diff
multi-line-comment := '/*' commented-block
commented-block := '*/' | (multi-line-comment | '*' | '/' | [^*/]+) commented-block
-slashdash := '/-' line-space*
+slashdash := '/-' node-space*

// Whitespace
```

zkat:
I definitely want the following to be legal:

```kdl
/-
my-node 1 2 3
```

so I’m not terribly inclined to change this at this stage

eilvelia:
Could you take a look at two small PRs I've sent (#441 and #442)?

tjol:
@zkat ckdl 0.2.1 is out now with opt-in support for KDL 2.0.0 as it stands. I'm just changing the defaults to KDL 2.0 now. Should be done in no time at all. Planning to call that version ckdl-1.0!

bgotink:
I've tagged version 0.2.0 of npm package `@bgotink/kdl` with KDL v2 support 🎉 ([release](https://github.com/bgotink/kdl/releases/tag/v0.2.0))

tjol:
No huge surprises in the process of changing the defaults, ergo:

__ckdl-1.0__ is released. This version supports both KDL v1 and v2; hybrid mode is the default for reading, and KDL v2 is the default for writing.

[Release](https://github.com/tjol/ckdl/releases/tag/1.0) / [Python package](https://pypi.org/project/ckdl/1.0/)

Feel free to mention in the main announcement.

zkat:
LFGGGGGGG

zkat:
Thank you everyone!! Great job!! We did it!!!

PR #435, opened 2024-12-15
Remove nonexistent equals-sign from the grammar

eilvelia:


zkat:
Aaaaa thank you. Nice catch

PR #437, opened 2024-12-16
add extra javascript implementation

bgotink:


zkat (approved):
oh exciting! another format-preserving one!!

PR #438, opened 2024-12-16
More tests for backslash behaviour

tjol:
Tests for the cases brought up by @eilvelia and @bgotink in #434, as well some more tests for (non-string) escline


tests/test_cases/input/escline_node.kdl
```diff
@@ -1,2 +1,3 @@
 node1
+\
```
tjol, line 2: Based on the filename I figured this test was missing something...

eilvelia:
Oh, I almost sent a similar PR :)

tests/test_cases/input/multiline_string_escape_newline_at_end.kdl
```diff
@@ -0,0 +1,4 @@
+node """
+    a
+   \
```
zkat, line 3: can you add the `_fail` equivalent to this (where `a` would end up indented less than the resolved prefix)

tests/test_cases/expected_kdl/multiline_string_double_backslash.kdl
```diff
@@ -0,0 +1 @@
+node "a\\ b\na\\b"
```
zkat, line 1: this makes me _uncomfortable_

tests/test_cases/input/escline_node.kdl
bgotink: This is allowed per spec, but I'm surprised by that.

`line-space` doesn't contain `escline`, only `node-space` has `escline`
All whitespace between nodes is `line-space`, so on first sight this test would be invalid, but&hellip;
`base-node` starts with `slashdash? type? node-space* string` which means this is valid

```kdl
node1
\
node2
```

but this isn't

```kdl
(lorem)node1
\
(lorem)node2
```

I personally always read (and implemented) the start of `base-node` as `slashdash? (type node-space*)? string`, as I interpreted the node-space to be meant as space between the type and name.

tests/test_cases/input/escline_alone.kdl
```diff
@@ -0,0 +1 @@
+\
```
tjol, line 1: Now I'm confused: is this actually allowed? I think it should be (just allow all whitespace to be escaped even if it's meaning less) but right now I'm not sure what the spec actually says right now

tests/test_cases/input/escline_node.kdl
zkat: ...I noticed this recently in kdl-rs too and it made me wonder whether `escline` _should_ be part of `line-space`.

tests/test_cases/expected_kdl/multiline_string_double_backslash.kdl
zkat: oh. Oh no it's fine. I misread it. For some reason I thought the string ended with a bell (`\b`). But no, this is fine.

tests/test_cases/input/escline_node.kdl
zkat: what in the spec allows this, btw?

tjol: Ah I see:
```
(type) \
node
```

is allowed and `(type)` is optional, therefore
```
\
node
```
is allowed.

Quite separately,
```
/- \
(type)node
```
and even
```
/-
\
(type)node
```
are allowed because `/-` can always be followed by `node-space`, even when not in a node.

IMHO, this is too confusing and `escline` belongs in `line-space`.

PR #439, opened 2024-12-16
reword interaction multiline + whitespace escape

bgotink:
Trying to resolve https://github.com/kdl-org/kdl/pull/434#discussion_r1886757547


zkat (approved):
I like this, but having others chime in would be good.

bgotink:
@tjol I gave it another try, could you take another look?

tjol:
I like it!

zkat:
I like this a lot, too! Thanks for taking the time!

PR #440, opened 2024-12-16
Make line-space a superset of node-space

tjol:
This is a proposal to simplify the `line-space` and `node-space` rules with the effect of allowing `escline` anywhere where whitespace is allowed.

This comes out of the discussion in #438.

The general idea (which I find intuitive) is that nodes can contain all whitespace _except_ unescaped newlines, because these would end the nodes. The space between nodes can simply contain all whitespace. Sure, `\` doesn't do anything, but so what? In the current grammar it is already allowed in a few select places where it doesn't do anything.


eilvelia:
I personally would very much support this. One of the things I didn't like about kdlv1 is that there's too many types of whitespace used in different contexts. Kdlv2 is better in this regard, and this change would make all whitespace rules (outside of strings) identical except for the non-newline requirement inside a node. It'll make parsing using a bottom-up parser easier (not requiring the lexer to emit escline as a token).

zkat (approved):

zkat:
I really really support this. Has this broken any of the tests for your parser in interesting ways?

tjol:
I'm not using the formal grammar directly in ckdl, so all I can say is that the way I understand this change is equivalent to the what ckdl does, and that passes all the tests in this repo (and in #438).

There are no tests in `release-2.0.0` right now that test that you _can't_ escape newlines outside of nodes, and one test that tests an edge case where you can (`slashdash_escline_before_node.kdl`).

zkat:
@tjol do you mind adding some tests to confirm that free esclines are valid? Or were they already there?

zkat:
oh duh nevermind that's what #438 was about lol. Sorry my head's all over the place.

tjol:
Still, can't hurt to add a few more!

Issue #443, opened 2024-12-21
Add an explicit marker for v2 to differentiate from v1 for non-programming parsers


eugenesvk:
This was originally raised in https://github.com/kdl-org/kdl/issues/372#issuecomment-1934373746, but is still unresolved. The suggestions offered in that discussion are only applicable to the **full-blown programming parsers** that can allow easy application of arbitrary logic with arbitrary "backtracking".

But for "simpler" cases like **syntax highlighting** it's not a viable option since you don't have that kind flexibility.

Case in point: the only syntax highlighting extension supporting v2 ([VSCode](https://github.com/kdl-org/vscode-kdl)) illustrates it - it doesn't support both v1 and v2, instead it suggests installing a separate extension for v1, but then which of the extensions would you apply to an actual file if there is no indicator???

Why not add something like `%YAML 1.2` as an optional indicator on the first line? Or maybe, in the unique spirit of KDL, the first slash-dashed node on the first line could signal version?
`/-v2` 
`/-v 2`
`/-kdl v=2`

zkat:
I really strongly don’t want this—I think it adds too much noise, and is ultimately unnecessary.

For things like syntax highlighters, all syntax highlighters I know of usually support a “heuristic” that you can provide to disambiguate files with the same extension. I haven’t done that with the VS Code one mostly because time, but it’s definitely a thing that should be done

eugenesvk:
> support a “heuristic” 

That's precisely what I'm asking KDL to add!!!

Here it is from Sublime https://www.sublimetext.com/docs/syntax.html#header

> Header
> The allowed keys in the header area are:
> 
> file_extensions ...
> 
> first_line_match
> When a file is opened without a recognized extension, the first line of the file contents will be tested against this regex, to see if the syntax should be applied.

There is no magic that would allow syntax highlighters to guess the syntax if the syntax supports nothing!

I mean, it's impossible to do in the general case even with the flexibility of programmable parsers: let's say my syntax highlighter is awesomely flexible and allows me to parse both and pick whichever version has no version-revealing errors (might be a weird editing experience where adding v2 syntax at the end of the file would suddenly conver the whole document to v2 and highlight errors in all the previosly valid doc which might not even be visible to the user).

[...]

PR #444, opened 2024-12-21
Add an optional version marker

eugenesvk:
As discussed in the linked issue, adding an **OPTIONAL HINT** for the specific KDL version

`/- kdl-version 2` (or `1`) with the only difference from a slash-dashed node is that only unicode space is allowed, so no newlines and (multiline) comments

Didn't update the v1 spec as it's described as an unchanging legacy document

Also didn't mention in the readme as the marker isn't considered important enough :)

Also didn't add any tests as the status is a hint

Close https://github.com/kdl-org/kdl/issues/443

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -926,7 +926,7 @@ line-space := node-space | newline | single-line-comment
 node-space := ws* escline ws* | ws+
 
 // Version marker
-version := '/-' unicode-space* 'kdl-version' unicode-space* ('1' | '2')
+version := '/-' unicode-space* 'kdl-version' unicode-space* ('1' | '2') unicode-space* newline
 ```
 
 ### Grammar language
--- a/SPEC.md
+++ b/SPEC.md
@@ -926,7 +926,7 @@ line-space := node-space | newline | single-line-comment
 node-space := ws* escline ws* | ws+
 
 // Version marker
-version := '/-' unicode-space* 'kdl-version' unicode-space* ('1' | '2') unicode-space* newline
+version := '/-' unicode-space* 'kdl-version' unicode-space+ ('1' | '2') unicode-space* newline
 ```
 
 ### Grammar language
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -920,6 +924,9 @@ newline := See Table (All Newline White_Space)
 line-space := node-space | newline | single-line-comment
 // Whitespace within nodes, where newline-ish things must be esclined.
 node-space := ws* escline ws* | ws+
+
+// Version marker
+version := '/-' unicode-space*? 'kdl-version' unicode-space*? ('1' | '2') line-space*
```
eilvelia, line 929:
```suggestion
version := '/-' unicode-space* 'kdl-version' unicode-space* ('1' | '2')
```

`?` and `line-space*` are redundant here

SPEC.md
```diff
@@ -920,6 +924,9 @@ newline := See Table (All Newline White_Space)
 line-space := node-space | newline | single-line-comment
 // Whitespace within nodes, where newline-ish things must be esclined.
 node-space := ws* escline ws* | ws+
+
+// Version marker
+version := '/-' unicode-space* 'kdl-version' unicode-space* ('1' | '2')
```
zkat, line 929: This should be followed by a line-space to make sure it’s always on its own line

zkat (changes requested):

SPEC.md
eilvelia: `line-space` doesn't require newlines. Perhaps it would be good to add `node-terminator` as in a regular node?

eugenesvk: `line-space` can be `node-space`, which can be `ws+`, which can be `unicode-space`, which doesn't mandate a newline before the node
So this should either be a `newline` or `unicode-space* newline`?

eugenesvk: Node terminator seems like a better idea!, though it allows `;`, so also doesn't mandate newlines?

zkat: yeah, I think it's gonna need to be `unicode-space* newline.

zkat (approved):

Issue #331, opened 2023-06-25
Vertical tab U+000B missing from the whitespace table


eugenesvk:
When comparing the list of whitespace with some regex engine predefined list of whitespace I've noticed that KDL spec table includes horizontal `U+0009`, but is missing vertical tabspace `U+000b`

Is this just a simple omission or is there a reason one whitespace tab does not belong?

https://www.compart.com/en/unicode/U+0009

https://www.compart.com/en/unicode/U+000b

tabatkins:
(Reverted, since this should have gone into the v2 branch as a technically-breaking change.)

zkat:
I'm actually gonna close this, since it's already in the v2 branch

PR #446, opened 2024-12-21
fix: move vertical tab to the line-breaking whitespace table

eugenesvk:
Currently linebreaking vertical tab is in the non-line-breaking whitespace table

(this mistake was prompted by https://github.com/kdl-org/kdl/issues/331)

Not sure what the impact of this is re. parsers and tests and other docs and whether this should be changeloged or just a documentation issue.

https://www.unicode.org/versions/Unicode13.0.0/ch05.pdf that the spec references agrees:

(bad copy&paste, blame PDF standard!)
5.8 Newline Guidelines
....
Table 5-1. Hex Values for Acronyms
Acronym Name Unicode ASCII EBCDIC
Default z/OS
CR carriage return 000D 0D 0D 0D
LF line feed 000A 0A 25 15
CRLF carriage return and
line feed <000D 000A> <0D 0A> <0D 25> <0D 15>
NEL next line 0085 85 15 25
VT vertical tab 000B 0B 0B 0B
FF form feed 000C 0C 0C 0C
LS line separator 2028 n/a n/a n/a
PS paragraph separator 2029 n/a n/a n/a

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -742,7 +742,6 @@ space](https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt):
 | Name                 | Code Pt |
 |----------------------|---------|
 | Character Tabulation | `U+0009`  |
-| Line Tabulation      | `U+000B`  |
 | Space                | `U+0020`  |
 | No-Break Space       | `U+00A0`  |
 | Ogham Space Mark     | `U+1680`  |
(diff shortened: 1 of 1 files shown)


eilvelia:
> Not sure what the impact of this is re. parsers and tests and other docs and whether this should be changeloged or just a documentation issue.

I think the `Vertical tabs (U+000B) are now considered to be whitespace.` changelog line should be changed to `Vertical tabs (U+000B) are now considered to be newlines.`

bgotink:
There's one test called `vertical_tab_whitespace.kdl` that would need to be updated.

https://github.com/kdl-org/kdl/blob/0e58f61b18380d62148fe55c7d906a9e69889f08/tests/test_cases/input/vertical_tab_whitespace.kdl#L1

https://github.com/kdl-org/kdl/blob/0e58f61b18380d62148fe55c7d906a9e69889f08/tests/test_cases/expected_kdl/vertical_tab_whitespace.kdl#L1

zkat:
Please go ahead and update the test. I think this is minor enough that we can assume most implementations will just work and they’ll eventually catch up with the change.

zkat:
nm I'll do it real quick. I already did this in kdl-rs :)

Issue #451, opened 2024-12-24
Add a test for escaped surrogates


eugenesvk:
https://github.com/kdl-org/kdl/pull/450

PR #452, opened 2024-12-24
test: add a fail-test for escaped non Unicode Scalar Values

eugenesvk:
Closes https://github.com/kdl-org/kdl/issues/451


zkat (changes requested):
We should probably have separate failing cases for each thing we intend to test, otherwise a single (correct) failure might obscure failures for other items

eugenesvk:
Thought about that, but that's solved via parameterized tests, then can properly test that each \u codepoint in the range fails, otherwise didn't want to create hundreds of separate files, and without that it's no help separating 5 codepoints currently in the text out

zkat:
I would prefer not to complicate the test suite right now. 4-6 tests at the appropriate boundaries should be “good enough”

zkat (approved):

PR #469, opened 2025-01-09
Identifier strings can technically be terminated by EOF

danini-the-panini:
The documents `node` and `foo bar` are valid documents, and the identifier strings "node" and "bar" here are terminated by EOF, so technically identifier strings can be terminated by EOF, unless documents always have to have trailing newline/whitespace.


zkat:
I guess it's been treated as kind of implicit that "termination" includes EOF, but I like having this more explicit. Thanks!

zkat:
Oops forgot to merge

Issue #475, opened 2025-01-12
self-expression: Add KDL grammar in KDL


eugenesvk:
Currently the grammar is only available as an unformatted blob of text in the markdown document.

In addition it would be great if KDL could express itself and you could read a properly formatted/highlighted grammar document

So that instead of this gray uniformity 
<img width="416" alt="Image" src="https://github.com/user-attachments/assets/2cb79fb4-fab2-434a-968d-e275233dda7a" />

you could have something like this
<img width="400" alt="Image" src="https://github.com/user-attachments/assets/51935e00-3a44-4b3d-9f1c-80a3b3cb9e2b" />

tabatkins:
It's not a "blob of text", it's a grammar description in a well-known grammar description language. Writing the grammar "in KDL" would require us to define a *new* grammar description language using KDL as the syntax base, which would be a project on its own. I don't see great value in that.

It would also require anyone reading the grammar that defines KDL to already know the grammar of KDL. (Which, tbf, is in broad strokes very simple, but still.)

(I'll also note that what you have written there, while technically a valid KDL document, isn't a *meaningful* KDL document. It treats section comments as nodes, and requires you to examine the exact order of properties and attributes, since a grammar term is a property *and the following attributes up to the next property*; for example, "keyword" is defined by its value (`boolean`) and the following two attributes (`|`, an ident string; and `#null`). No one would actually define a KDL language like that.)

eugenesvk:
> It's not a "blob of text", it's a grammar description in a well-known grammar description language

Which one and why isn't it marked as such in the spec? The description implies ABNF, but when I used ABNF syntax highlighter, it failed since apparently `//` isn't a comment there. It also failed on char ranges and some other stuff. So I doubt current description is really specced to a "well-known language"

> I don't see great value in that.

The great value is that it's not a grey blob of text, but a syntax-highlighted doc, which improves readability

> KDL to already know the grammar of KDL

Not really, you don't need to understand that `\` is line continuation, just read in the grammar that `\` is not part of grammar, but part of the extra syntax. And this isn't that different from reading in the current `Grammar language` that explains the basics such as

> Single quotes (`'`) are used to denote literal text


>It treats section comments as nodes

it treats sections as nodes, comments are args, but more imporantly

>, and requires you to examine the exact order of properties and attributes,

yes, it requires (and is designed for) READING, not programmatic parsing

> meaningful KDL document. 

so it is meaningful for reading, hence instead of
[...]

PR #476, opened 2025-01-12
Add KDL2 grammar in KDL in examples

eugenesvk:
I'm adding a KDL2 grammar in KDL2 that I used as a reference when updating the syntax highlighting package. It's not a literal translation of the grammar:
- it add parent-child relationships via indentation (`document = bom? version? nodes` then `version`/`nodes` are +1 indent vs. document)
- it's more condensed: e.g., `node-space+` takes way too much space for a reference to space, so I use `n␠+` instead. Also some comments are inlined to not visually mix rules with description
- at the same time it's more explicit in some cases, e.g., it has all the non-line breaking whitespaces explicitly, to avoid the need to go back&forth between reading the spec and reading the grammar
- it uses the power of unicode to avoid quotes, e.g., using a 1 symbol `＃` instead of 3 `"#"` and to be more expressive (`char🛑` signals invalid chars, `str🆔` signals bare identifiers), etc.

And in the end you get a proper syntax-highlighted reading experience

KDL grammar 1

KDL grammar 2

partially addresses https://github.com/kdl-org/kdl/issues/475

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -837,7 +837,8 @@ except for non Unicode Scalar Value, which can't be represented even as escapes)
 
 This is the full official grammar for KDL and should be considered
 authoritative if something seems to disagree with the text above. The [grammar
-language syntax](#grammar-language) is defined below.
+language syntax](#grammar-language) is defined below. For a version converted to KDL,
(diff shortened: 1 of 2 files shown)


tabatkins:
Per discussion in #425, we're not interested in pursuing this.

PR #483, opened 2025-01-16
Support empty multiline strings

tabatkins:
Per <https://github.com/kdl-org/kdl/discussions/481>, the grammar doesn't actually allow the first and last newline in a multiline string to be the same (indicating an empty string). This looks like the minimal fix for it.


zkat (approved):

zkat:
Can you write a super simple test specifically for this?

tabatkins:
Done, wrote a handful because why not.

PR #484, opened 2025-01-17
minor tweaks to wording/clarifications

zkat:
these are just a few minor wording tweaks and a test I thought might be useful, since it occurred to me that we can do stuff like spread a big base64 binary across several lines using `\`.

I also snuck in `Ascii85` because that seems handy to mention, but I'm happy to take it out if that's controversial.


danini-the-panini (approved):

tabatkins (approved):

zkat:
yeah T_T

PR #486, opened 2025-01-17
KDL Schema Language v2

### LINKED ISSUE #504: New KDL Schema
this is the tracking issue for the ongoing work on the new KDL Schema language

zkat:
This is an updated version of the KDL Schema modeled after a combination of JSON Schema and XSD. This PR should be used for general discussion of the spec as we edit it, and suggestions can be made in-thread or in the form of PRs against the `zkat/schema-v2` branch it's based on.

Most of the work on the spec is happening in the `schema/ksl-schema.kdl` file, which is the KSL Schema for the KSL Schema Language itself, which will eventually be translated to a more formal prose spec in the spec markdown itself. Please ignore the current spec file, as it is out of date.

## Status

* [x] New baseline for v2 schema language
* [ ] Make sure `@ksl:schema` can be disabled
* [ ] Do a first-pass review of the current draft
* [ ] Something like JSON Schema's [`dependentRequired`](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentRequired)
* [ ] Something like JSON Schema's [`dependentSchemas`](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentSchemas)
* [ ] Something like JSON Schema's [`If-Then-Else`](https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse)
* [ ] Something like JSON Schema's [composition keywords](https://json-schema.org/understanding-json-schema/reference/combining) (`anyOf`, `allOf`, `oneOf`, `not`). This might also be something we do using an API like XSD's `<xsd:group>` and associated element types.
* [ ] some utility mechanism(s) to help feed information into LSP semanticTokens for folks who are doing DSLs, so things will be automatically highlighted with those special classes, using schema information.
* [ ] Translate `ksl-schema.kdl` to a proper prose spec doc (wait a bit)


LHolten:
How do you specify that a property is "nullable" (for example a property that is allowed to be a date-time or null)?

zkat:
@LHolten `type string #null`

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
+@kdl:schema "https://github.com/kdl-org/kdl/blob/main/schema/kdl-schema.kdl"
```
TheLostLambda, line 1: Where will this `@kdl:schema` be used? If it doesn't have to be a reachable URL, then do you still think that it's valuable to have as a top-level key like this?

Do you envision KDL documents adding this to their top-level? Does that mean users are expected to handle this node when parsing with `kdl-rs`? Or should KDL parsers remove this node from the AST before passing it on to user code? If so, then I think that would require a change to the KDL spec? 

I'm not opposed to this, though I'd personally probably keep them separate and give any sort of validator two files (the data and the schema it follows) — unless I'm misunderstanding though, I think this might be better as a field in `metadata`.

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
...
+
+metadata {
+    // TODO: update this link when we're ready to release something.
+    link "https://github.com/kdl-org/kdl/blob/main/schema/cargo.kdl" rel=self
```
TheLostLambda, line 5: Likely because I'm super new to this, but what's the value of `rel=...` here?

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
...
+    title "Cargo Schema" lang=en
```
TheLostLambda, line 6: For the `lang` stuff, I can see the value of having that as a property so that you can easily have mixed-language documents, but would be be useful to just have a `metadata.lang` node that sets the document default unless otherwise specified?

Also, how does one specify the language of a string in a property? Like when you have:
```kdl
node version title="The version of the package."
```
How do you mark the lang of the `title` string?

Again likely my cluelessness: do many other schemas have something like this?

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
...
+    description "KDL-based translation of the Cargo.toml schema." lang=en
+    author "Kat Marchán" {
+        link "https://github.com/zkat" rel=self
+    }
+    link "https://github.com/kdl-org/kdl" rel=documentation
```
TheLostLambda, line 11: Maybe the `rel` could be changed to something much more general, like `desc` for any string description of the link?

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
...
+    link "https://doc.rust-lang.org/cargo/reference/manifest.html" rel=documentation
+    license "Creative Commons Attribution-ShareAlike 4.0 International License" spdx=CC-BY-SA-4.0 {
+        link "https://creativecommons.org/licenses/by-sa/4.0/" lang=en
```
TheLostLambda, line 14: I think this is a really good use of `lang`! It's possible to imagine software that has a link for license or documentation but the exact URL depends on the language your computer is configured to use (if the website doesn't automatically reroute you)!

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+children {
+    node package title="Describes a package" {
```
TheLostLambda, line 19: Personally I'd swap `title` out for `help`, `desc`, `info` etc — it seems like more of a description than the title of a node? I suppose `title` feels like it overlaps with name to me?

schema/cargo.kdl
```diff
@@ -0,0 +1,172 @@
...
+        children {
+            node name title="The name of the package" {
+                required
```
TheLostLambda, line 22: Elsewhere I've also seen `optional` — by default is everything described required? Or optional?

To me it would make sense maybe that by default every described thing is required unless you mark it optional?

schema/kdl-schema.kdl
```diff
@@ -0,0 +1,477 @@
...
+    node @kdl:schema title="Schema URL" description="URL for the schema describing this document. It MUST be a properly-formatted URL. Implementations MAY attempt to visit it, but MUST NOT assume it is valid." {
+        arg {
+            type string
+            format url irl
+        }
+    }
+    node metadata title="Schema metadata" description="Contains metadata about the schema itself." {
+        min 1
+        children {
+            node id title="Schema identifier" description="The unique identifier for this schema. MUST be a valid URL/IRL. Implementations MAY attempt to visit it, but MUST NOT assume it is valid." {
```
TheLostLambda, line 40: Ah! Here there is a `title` and `description`! I see!

Hmm, just looking at this makes me maybe think they could be moved to be children? Just because the strings have a tendency to get long and it would be a bit easier on the eyes if that long content were inside of the `{}` in my opinion?

TheLostLambda: Something very silly (not sure how I feel about it) if it were a child it could have short and long under one `description`?

```kdl
description "Schema identifier" long="The unique identifier for this .."
```

TheLostLambda (commented):
Hi @zkat !

Sorry this is a super erratic and incomplete review of things, but I figured I should share what I had so far before coming back to it! Lots of the comments are likely just me not being super familiar with schemas or missing something, but hopefully a few are interesting?

My plan for tomorrow is to write some schemas for some of my KDL files, bending this syntax if needed to get something that feels clean and minimal, then I'll hopefully have an idea of where things can be tweaked!

The only other specific thing I can think is related to #492 somewhat — I personally think arbitrary links around documents might be a little wacky, and all of my files where I've been "inheriting" properties have looked like:

```kdl
residues {
    types {
        Monosaccharide {
            functional-group "Hydroxyl" at="Reducing End"
            functional-group "Hydroxyl" at="Nonreducing End"
            functional-group "Hydroxyl" at="6-Position"
            functional-group "Acetyl" at="Secondary Amide"
        }

        AminoAcid {
            functional-group "Amino" at="N-Terminal"
            functional-group "Carboxyl" at="C-Terminal"
        }
    }

    Monosaccharide "g" "N-Acetylglucosamine" {
        composition "C8H15NO6"
    }
    Monosaccharide "m" "N-Acetylmuramic Acid" {
        composition "C11H19NO8"
        functional-group "Carboxyl" at="Lactyl Ether"
    }
    Monosaccharide "x" "Unknown Monosaccharide" {
        composition null
    }

    AminoAcid "A" "Alanine" {
        composition "C3H7NO2"
    }
    AminoAcid "B" "Diaminobutyric Acid" {
        composition "C4H10N2O2"
        functional-group "Amino" at="Sidechain"
    }
    AminoAcid "C" "Cysteine" {
        composition "C3H7NO2S"
    }
    AminoAcid "D" "Aspartic Acid" {
        composition "C4H7NO4"
        functional-group "Carboxyl" at="Sidechain"
    }
    AminoAcid "E" "Glutamic Acid" {
        composition "C5H9NO4"
        functional-group "Carboxyl" at="Sidechain"
    }
    AminoAcid "F" "Phenylalanine" {
        composition "C9H11NO2"
    }
    AminoAcid "G" "Glycine" {
        composition "C2H5NO2"
    }
    AminoAcid "H" "Histidine" {
        composition "C6H9N3O2"
    }
    AminoAcid "I" "Isoleucine" {
        composition "C6H13NO2"
    }
    AminoAcid "J" "Diaminopimelic Acid" {
        composition "C7H14N2O4"
        functional-group "Amino" at="Sidechain"
        functional-group "Carboxyl" at="Sidechain"
    }
// SNIP ...
```
Where I have some node "hard-coded" to store some templates that are used in the rest of the document, so the "variables" are all in one place and you know where to scroll to to find that information. Now, I can see how that might be tricky when you're writing something like a schema for the schema, and you need a lot more linking? If that's the case, then maybe giving relative or absolute node "paths" in something like KQL could be good? I'll have to think a bit more on it, but I think tagging one node, then linking to that tag from anywhere could be a bit confusing and tricky to maintain / keep the ID's always in sync.

Sorry, I've rambled far far too much here, but as a concluding thought, a bit like the unicode quotes "go big or go home" discussion, I'd lean on the side of keeping things more general, minimal, and future-proof personally — being a bit cautious of feature-creep. Things like `rel=self|documentation` or author ORCIDs being in the specification feel like things that be over-specifying? Though that's just my gut reaction!

I'll have much more practical and hopefully helpful stuff to share after I try writing a couple of schemas myself! Sorry in the meantime and hopefully there is a nice regex for finding the useful needle in this haystack (like that insane semver one) :sweat_smile:

schema/cargo.kdl
bgotink: This is similar to JSONSchema's `$schema`, which points to a valid (relative) URL.

This is not required, e.g. I can edit my `angular.json` and get LSP support because the Angular VSCode plugin interprets any file called `angular.json` as an Angular configuration document.
But, for schemas that don't come with editor plugins or that don't have a filename convention, it is very handy that you can add that `$schema` property and the JSON plugin will kick in and provide LSP support.

schema/cargo.kdl
zkat: Oh I changed this field to `id`, I just haven’t pushed that change I guess. That’ll be clearer (and not need rel)

schema/cargo.kdl
zkat: I’m ok removing the lang stuff tbh.

schema/cargo.kdl
zkat: title is what jsonschema uses so I just went with that because it seemed as good as anything else. It WILL largely be used as the “title” for these items when displayed by editors and doc generators

schema/cargo.kdl
zkat: Some things are optional, some are required: any `arg` specified is required unless marked optional (which aligns with people generally expecting the value in a key/value pair to always be present). For nodes and properties, optional is the default because people generally don’t expect to have every single field itself be required. It might seem confusing but I think the ergonomics pan out when you think about actual usage:

in JSON, most of the time, only a small number of possible properties are required in an object. But if you have a property, you always have to have a value (you could argue that missing args should be null, but I don’t like playing that game as much as I can avoid it)

schema/kdl-schema.kdl
zkat: Yeah that could work. They could also be both. I do that in a few places

schema/cargo.kdl
TheLostLambda: Okay, I can definitely see that as nice to have for editor / tooling support! Would implementations like `kdl-rs` strip this out of the AST returned to users? Essentially just because I agree it would be something nice to add to get LSP support, but I wouldn't want that addition to affect any result of parsing!

Is this the sort of thing that could / should be a standard-formatted comment? Like how sometimes people seem to set vi / vim settings in comments with `vi:` or whatever at the top of a file?

schema/kdl-schema.kdl
TheLostLambda: I'd also not be opposed to having it in both places — I've done that in some of my KDL files before too! Though definitely worth making sure that whatever the final solution is that it's not introducing more complexity than it's worth / that it's relatively consistent so it stays intuitive when a property can become a child :)

schema/cargo.kdl
TheLostLambda: I think that's very fair — I think args as default required is good, and same with properties as default optional! I could see nodes going either way, perhaps, but I think I'm biased by the KDL configs I've been creating.

Now that I understand that behavior, I'm perfectly happy with it! Thanks for your patience and explanations!

schema/cargo.kdl
zkat: Users would have to ignore this field, if they want it to be present. Just like folks are free to ignore `$schema` when using JSONSchema.

I thought about using a comment for this, but that seems... somehow worse? I don't know. I feel like it should just be data.

schema/cargo.kdl
TheLostLambda: That's fair — I suppose I'd have to see more examples, though intuitively I might still lean in the `info` direction if it's the sort of thing going into an LSP or doc generator! Intuitively I read that as the title being `package`, with docs something like:

```md
# `package`
Describes a language

Arguments: ...
Properties: ...
```

Maybe even could be `doc` for documentation? That might be my favorite since it's clear it's meant to be for humans and isn't some sort of ID to be referred to. But ultimately that's just a small bikeshedding matter and I'll be happy however it ends up! :)

schema/cargo.kdl
TheLostLambda: I see! That makes more sense — `rel` as in RELationship to the document.

I suppose I'd still feel a little weird about using `enum` to try to capture every possibility there — maybe like the `(type)` syntax in KDL, we carve out a few common / obvious ones, so implementations can rely on those having fixed meanings, then leave the field open to any arbitrary strings for implementations to assign meaning to? This is probably my favorite solution; something like: https://github.com/kdl-org/kdl/blob/main/draft-marchan-kdl2.md#reserved-type-annotations-for-strings

The `enum self documentation` I spotted somewhere may just be incomplete, but I definitely think we'd want to leave it more open than just those two!

schema/cargo.kdl
TheLostLambda: I'll have a bit more of a ponder about it, but I agree that might be a good way to simplify things a little bit, since making it consistent would perhaps be a lot of work

TheLostLambda:
I can't seem to resolve review comments I've opened, so feel free to resolve and hide any of those threads that feel resolved!

schema/cargo.kdl
TheLostLambda: That's fair... I'll sleep on it — I'd personally find it a little wacky to have to describe in my schema for a KDL format the node that will be used to refer back to that schema. So, if I wanted to do what @bgotink mentioned, and link my current KDL file to a schema for editor support, I'd need to make sure that that schema includes the `@kdl:schema` node in it as well? If it's even a URL linked to a file I can edit?

But, before I whinge any more, I'll read up properly on `$schema` in JSON and do some proper thinking!

schema/cargo.kdl
TheLostLambda: My last question for now would then be what specifically `rel=self` would mean here? And if `rel=self` would be implied if you left off the `rel` property entirely.

zkat: I'm thinking I should add an `allow-others` to `enum`: this would still be useful because, while it would allow basically anything, the enum can still be used by editors to provide "known" completions.

In the case of `rel=self` in the `author` node, it means "a link related to the author", if in the metadata node, it would mean "a link related to the schema", and so on.

schema/cargo.kdl
zkat: We could do that Git Thing, where your first line is the "title", and everything else is the "details".

So you could do:

```kdl
node package description="""
   Describes a language

   Here are more details about how you can use this.
   Have you considered using KDL for all your configuration needs?
   """ {
   arg description="arg1\n\nsomething about it"
   prop foo description="Foo\n\nA fooish property"
}
```

And then, tooling would be able to strip that first line and generate a markdown like this:

```md
# `package` - Describes a language

Here are more details about how you can use this. Have you considered using KDL for all your configuration needs?

## Arguments

* arg1 - something about it

## Properties

* Foo (`foo`) - A fooish property
```

...or however you choose to format that data.

schema/kdl-schema.kdl
zkat: yeah I'm trying to be mindful of minimizing what I put into properties/arguments, and sometimes it makes sense to have them in both (for example, if you want to be able to provide multiple values for a property, or if having it in the main node line would cause a lot of noise vs nesting)

schema/cargo.kdl
TheLostLambda: I like this — it seems a lot more in-line with how most programming languages do their documentation too! It simplifies the node structure a bit and also makes use of KDLs really nice multi-line, deindented strings!

And I think `description` or `doc` / `documentation` are good names too :)

schema/cargo.kdl
zkat: I've changed this so you no longer need to remember to include this in your own schemas: it is now a _default_ item for all schemas, and can be disabled by doing `node @kdl:schema { undefine }`.

bgotink: > So, if I wanted to do what @bgotink mentioned, and link my current KDL file to a schema for editor support, I'd need to make sure that that schema includes the @kdl:schema node in it as well?

There are multiple versions of JSONSchema. The `$schema` in the schema is used to define the version of JSONSchema in which it is written, so it serves an extra purpose on top of enabling your editor to provide support while writing schemas.

For example, data file "my-data.kdl" has `@kdl:schema "./my-schema.kdl"`, schema file `my-schema.kdl` would then have `@kdl:schema "https://kdl.dev/schema/v0.2.kdl";` (fictitious URL)

> I thought about using a comment for this, but that seems... somehow worse? I don't know. I feel like it should just be data.

Agreed, both because it feels bad and because it's actually useful data. Just like the version of JSONSchema, this could be used to differentiate between different versions of configuration file formats. If this were a comment, you'd have to have a comment to a versioned URL ánd a version indicator in the data.

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,617 @@
...
+                                    but MUST NOT fail.
+                                    """ {
+                                    repeatable
+                                    arg title="ID @ksl:reference to another node" {
+                                        type string
+                                        optional
+                                    }
+                                    prop path title="KPath @ksl:reference to another node." {
+                                        type string
+                                        format kpath
```
bgotink, line 250: I assume this means we have to finalize the KPath 2.0 spec before we can finalize this spec?

zkat: yep. And possibly make a small modification to the v2 spec itself to add `kpath` as a "reserved" string format.

TheLostLambda:
Looks like lots of exciting updates — I'll aim to take a look this evening!

schema/cargo.kdl
TheLostLambda: Okay, I've done some pondering and perhaps come to the conclusion that the first draft of this was best, but with some caveats. If I'm understanding correctly, it sounds like there's two, slightly distinct, reasons to want something like this:

1. It's a way to link files to a schema that can be used to enrich editor / LSP-type support
2. It can also be used as a sort of idiomatic / canonical version field, if a format's schema has changed over time

But one of my problems remains if the `@kdl:schema` must be an uncommented node: say I'm working on a certain type of KDL config file — it's the type of thing that's complex enough that I'd like my LSP to know the schema and validate my edits. The issue, however, is that this config file is read by some code that looks like this:

```rs
#[derive(Serialize, Deserialize)]
#[serde(deny_unknown_fields)]
struct KdlConfig {
    bonds: BondData,
    modifications: ModData,
    // Etc...
}
```

Where the important part is the `#[serde(deny_unknown_fields)]` — which is think is a nice thing to have if you want to enforce that config files are free of unrecognized sections, something that could help a lot in catching user typos. Now I'm in a bit of a bind because, once again, I have to switch between 1) having `@kdl:schema` uncommented, so LSP support works but the config can't actually be loaded, or 2) commenting out that node, losing LSP support, but actually being able to test the config with the program it's meant for.

Obviously, however, this is a non-problem if the program already uses `@kdl:schema` as a standard convention for specifying a version:

```rs
#[derive(Serialize, Deserialize)]
#[serde(deny_unknown_fields)]
struct KdlConfig {
    #[serde(rename = "@kdl:schema")
    version: Url
    bonds: BondData,
    modifications: ModData,
    // Etc...
}
```

Then leaving `@kdl:schema` uncommented 100% of the time is perfectly fine!

To address both of these cases, we could:

1. Make it clear that `@kdl:schema` is the idiomatic way to version KDL formats, and set a good example by having the `kdl-schema.kdl` define the optional `@kdl:schema` node
2. Make sure it's clear to all LSP implementations that schema information can come in one of two forms: either an uncommented `@kdl:schema` node, I suppose anywhere in the file, or a `/- @kdl:schema` comment at the top of the file for when you want editor support, but the KDL format you're writing disallows it.
3. Maybe revert the "it is now a default item for all schemas" change? I think I'd prefer to reduce the amount of magic, and if a format wants to use `@kdl:schema` to keep track of versions, then I think it should say so explicitly.

In summary then, I think rolling this back to how it started (schemas must explicitly allow `@kdl:schema`) and making it clear that language servers should recognize both `@kdl:schema` and `/- @kdl:schema` (maybe allow other commented forms?) would be my ideal solution.

That way even if someone's program is explicitly disallowing unknown fields (to save users from typos) and doesn't care to version files, I can still get editor support.

Let me know what you think about demystifying the `@kdl:schema` node and "downgrading" it to something that's idiomatic convention / recognized by tooling.

SIDENOTE: Serde support for KDL will need to make decently heavy use of `#[serde(rename = "...")]`, and `quick-xml`, for example, already uses the "@..." prefix for attributes. I was thinking perhaps we could use "$..." for properties (since they felt kind shell-variable-ish), but we'll likely need at least one other special prefix reserved. If we used `@` for that, like `#[serde(rename="@arguments")]` or whatever, then reading `@kdl:schema` would become a bit more awkward (though not impossible, I don't think?). Might be a bit bike-sheddy, but we should perhaps think about either removing the `@`, or rule it out as a potential `serde` prefix!

Sorry for the flip-flop on this, but I think this is a solution I'd be 100% happy with <3

schema/cargo.kdl
TheLostLambda: After more thought, I think removing this might be best for the sake of keeping things simple and consistent!

schema/cargo.kdl
TheLostLambda: Though I'd have to do some thinking on the exact syntax, I 100% agree that conceptually there should be something like an `allow-others` for enums for exactly the reason you gave :)

A more general solution (I've not yet looked at your schema updates, so ignore me if you've already done something like this, I think we should have some sort of "sum-type" or union syntax. Maybe something like:
```kdl
one-of {
    string
    enum self documentation
}
```

Or something like that — it would mean less magic in `enum`, and also let you have places where either a number or string are acceptable with the same mechanism!

schema/ksl-schema.kdl
TheLostLambda: Do you mean as a type annotation? Like these? https://github.com/kdl-org/kdl/blob/main/draft-marchan-kdl2.md#reserved-type-annotations-for-strings

Something like `(kpath)"<whatever>`?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+            node version about="""
+                Schema semver version
+
+                The version number of this version of the schema, in semver
+                format.
+                """ {
+                arg about="Semver version number" {
+                    type string
+                    // https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string.
+                    pattern #"^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"#
```
TheLostLambda, line 179: Is this worth adding to `format` / `validate-as` and KDLs reserved type annotations? https://github.com/kdl-org/kdl/blob/main/draft-marchan-kdl2.md#reserved-type-annotations-for-strings

Personally, I can imagine more KDL documents with Semver strings than some of the other types already reserved! If you're already wanting to add something like `kpath`, then I reckon this would be helpful and would save other people the research and copy-paste of that regex beast!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                        date-time date time duration decimal currency country-2 \
+                        country- country-subdivision email idn-email hostname \
+                        idn-hostname ipv4 ipv6 url url-reference irl \
+                        irl-reference url-template regex uuid kpath \
+
+                        // Number validations
+                        i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 isize usize f32 \
+                        f64 decimal64 decimal128 \
+                        {
+                        allow-others
```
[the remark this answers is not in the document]
TheLostLambda: I've also just seen `allow-others` as a node inside of `children`, which makes perfect sense. I'll have to think more on this then...

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+        schema. They will be inert (that is, not directly apply to the document)
+        unless referenced by another node inside `document`.
+        """ {
+        children {
+            allow-others
+        }
+    }
+    node document {
+        ref children-node
+        default {
```
TheLostLambda, line 197: As I mentioned in my (too long) reply about this, I think I might prefer to not have the concept of `default` nodes... Though perhaps that's just because I'm struggling to see a use for them outside of making schema-schemas?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+            node @ksl:schema about="""
```
TheLostLambda, line 198: I do prefer the `kdl` to `ksl` change though!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                nodes are present, ALL schemas MUST successfully validate in
+                order for the document to validate, unless `warn` is `#true`.
+
+                In such a case, implementations SHOULD warn that validation has
+                failed and report which schema failed to pass--they SHOULD
+                include more details about what the specific failure was, but
+                MAY simply indicate that certain schema(s) failed to validate.
+                """ {
+                repeatable
+                prop warn {
```
TheLostLambda, line 215: Perhaps `warn` would be a bit clearer as `warn-only` or `allow-errors` or something similar — just `warn` sounds like setting it to `#true` would *increase* the strictness somehow, but perhaps that's just me reading `warn` strangely!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                Children are only allowed on nodes (or the toplevel document) if
+                at least one `children` node is present in their definitions.
+                """ {
+                children {
+                    node names about="""
+                        Child node name validations
+
+                        String validations to apply to all node names in this scope.
+                        """ {
+                        ref string-validations
```
TheLostLambda, line 240: Also, after coming across `ref` a number of times, I don't know if it's quite communicating what it does super clearly — how would you feel about `include ...` or something like that?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                    node allow-others about="""
+                        Allow other children
+
+                        If present/`#true`, allows child nodes in this scope
+                        other than the ones explicitly listed and those allowed
+                        by `names`.
+                        """ {
+                        arg {
+                            type boolean
+                            default #true
```
TheLostLambda, line 253: Could be me still wrapping my mind around schema-schema, but I don't know about this sort of `default` either — defaults feel like they should be the domain of the code loading the KDL?

I suppose having defaults in the schema like this implies that the validation step would actually *change the parsed AST* during validation to insert that `#true` if the value were initially missing, and I think that would be quite unexpected.

Maybe this is just best marked as `optional`?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                        ref about-mixin
+                        repeatable
+                        arg about="Node name\n\nThe name of the node." {
+                            type string
+                        }
+                        prop id about="Node identifier\n\nA schema-unique ID/anchor for this node." {
+                            type string
+                        }
+                        children {
+                            node ref about="""
```
TheLostLambda, line 271: As before, I'd maybe rename this `include` or similar!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                                Any children of this node will be merged into the node
+                                definition itself, as if written directly into the node.
+                                Any conflicting definitions will be overriden by
+                                definitions within the rest of the node.
+                                """ {
+                                children {
+                                    allow-others
+                                }
+                            }
+                            node undefine about="Undefine a node with this name" {
```
TheLostLambda, line 364: If we get rid of default stuff, can we remove this too?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                                        repeatable
+                                        arg {
+                                            type string
+                                            format kpath
+                                        }
+                                    }
+                                }
+                            }
+                            // TODO: be more precise and clear about merge + override behavior.
+                            node default about="""
```
TheLostLambda, line 352: Maybe it's worth deciding if KSL is about just syntax, or also semantics?

Personally I'd like it to just be something for describing what is syntactically valid without having any bearing on the semantics of the document — more concretely, I think the AST produced by the parser should be produced *first* and be read-only for the validation step.

Including default values I think blurs that line and means the same document would resolve to different ASTs with and without the schema!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                                optional. Including this option will require
+                                that this node always appear in its parent's
+                                children block.
+                                """ {
+                                arg {
+                                    type boolean
+                                    default #true
+                                }
+                            }
+                            node repeatable about="""
```
TheLostLambda, line 384: I don't know why I have mixed feelings about this name, and I don't know if I can think of anything I like more, but in case you have an opinion, the closest I can get to happy is `allow-multiple`?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                                        arg {
+                                            type boolean
+                                            default #true
+                                        }
+                                    }
+                                }
+                            }
+                            // TODO: add a feature that will let us specify that `args`
+                            // MUST be after any existing `arg` nodes in the current
+                            // scope. i.e. you can't do `node x { args; arg }`
```
TheLostLambda, line 584: Could be a child node like `after` or `before` with a reference?

So in `args` it could be:
```kdl
node args {
    after super.arg
}
```

For example!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                repeatable
+                args {
+                    min 1
+                    type string
+                    enum string boolean number integer #null
+                    distinct
+                }
+            }
+            // TODO: establish equality expectations.
+            node const about="""
```
TheLostLambda, line 823: Maybe if we rename `enum`, then can get rid of this?

Maybe if `enum` were `has-value` (hmm, perhaps that name too can be improved...), it could be:
```kdl
has-value 2
has-value self documentation
has-value allow-others=#true u8 f64 date // Etc...
```

Overall, I think having, where possible, one right way to right things would be a good property and keep things simple!

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                        cases where there's a set of suggested values, but
+                        others are acceptable. This information can then be used
+                        by tooling to e.g. suggest completion items.
+                        """ {
+                        arg {
+                            type boolean
+                            default #true
+                        }
+                    }
+                    node - about="Enumeration choice" {
```
TheLostLambda, line 858: A bit confused by this `-` here? I may have just missed it elsewhere?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                        in the node.
+
+                        While apparently redundant, this option may be useful in
+                        cases where there's a set of suggested values, but
+                        others are acceptable. This information can then be used
+                        by tooling to e.g. suggest completion items.
+                        """ {
+                        arg {
+                            type boolean
+                            default #true
```
TheLostLambda, line 855: Even if we don't remove defaults, as I think we should, I think this should be `#false`?

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+
+                Sets a default value when optional. That is, it requires
+                `optional` for `arg` nodes, and doesn't do anything useful if a
+                `prop` is marked `required`, though it is not invalid to do so.
+                """ {
+                arg
+            }
+        }
+    }
+    node about-mixin id=about-mixin  {
```
TheLostLambda, line 889: I've maybe shifted my opinion a bit vs the past — in line with, ideally, "one right way", maybe we should be a lot more hesitant to allow things as both properties and children.

TheLostLambda (commented):
I still didn't quite have time to pour over every aspect in detail, but as before I hope some of the comments are helpful!

Feel free to resolve threads liberally to keep things manageable :)

TheLostLambda:
Also something you may have already thought about, but a massive part of the KDL validation I've been doing is maybe something I saw a comment referring to "cross-validation" about?

In a shortened version of the example I already shared in this thread:
```kdl
residues {
    types {
        Monosaccharide {
            // SNIP ...
        }

        AminoAcid {
            // SNIP ...
        }
    }

    Monosaccharide "g" "N-Acetylglucosamine" {
        composition "C8H15NO6"
    }

// SNIP ...

    AminoAcid "A" "Alanine" {
        composition "C3H7NO2"
    }

// SNIP ...
```

All of the children in the `residues` section have to have node-names that match one of the `types` in `residue.types`!

That would be a super helpful validation to have for schema where there is some sort of definitions section!

TheLostLambda:
Finally (for today), is the plan to have a working validator implementation before stabilizing the schema? It would be very very helpful to actually test writing some real schemas, and I'm sure would be invaluable for making sure the `ksl-schema.kdl` file is valid before it's stabilized!

schema/ksl-schema.kdl
zkat: It might seem confusing, but the logic is something like:

1. no `allow-others` node: the value is false, because there’s no value at all
2. An `allow-others` node with no argument: the value is `#true`, because we have the node, then look up the value, the value is missing, therefore we use the default
3. `allow-others #false`: forces the value to false.

the reason for doing things this way is that so nodes that work this way “present or not” can be overridden when using `ref`. So if you’re “inheriting” a node that includes `allow-others`, you can turn off that default

zkat:
> Also something you may have already thought about, but a massive part of the KDL validation I've been doing is maybe something I saw a comment referring to "cross-validation" about?
> 
> 
> 
> In a shortened version of the example I already shared in this thread:
> 
> ```kdl
> 
> residues {
> 
>     types {
> 
>         Monosaccharide {
> 
>             // SNIP ...
> 
>         }
> 
> 
> 
>         AminoAcid {
> 
>             // SNIP ...
> 
>         }
> 
>     }
> 
> 
> 
>     Monosaccharide "g" "N-Acetylglucosamine" {
> 
>         composition "C8H15NO6"
> 
>     }
> 
> 
> 
> // SNIP ...
> 
> 
> 
>     AminoAcid "A" "Alanine" {
> 
>         composition "C3H7NO2"
> 
>     }
> 
> 
> 
> // SNIP ...
> 
> ```
> 
> 
> 
> All of the children in the `residues` section have to have node-names that match one of the `types` in `residue.types`!
> 
> 
> 
> That would be a super helpful validation to have for schema where there is some sort of definitions section!

Cross validation is in the todo list

zkat:
> Finally (for today), is the plan to have a working validator implementation before stabilizing the schema? It would be very very helpful to actually test writing some real schemas, and I'm sure would be invaluable for making sure the `ksl-schema.kdl` file is valid before it's stabilized!

Yes I'm planning on starting an implementation soon

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                        ref metadata-link
+                    }
+                }
+            }
+            node contributor about="""
+                Schema contributor
+
+                A contributor to the schema might not be considered an author.
+                """ {
+                ref metadata-author
```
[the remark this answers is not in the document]
zkat: Remember that this is inheriting the format, not the data.

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                    type string
+                }
+            }
+            node min-length about="""
+                Minimum string length
+
+                Minimum length of the value, if it's a string.
+                """ {
+                arg {
+                    gte 0
```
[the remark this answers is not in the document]
zkat: I think it's important for the "mathy" validators to be clearly related to one another. `max`/`min` mean exactly one thing right now (related to child count). I think having the lt/lte/gt/gte family has a lot of value

schema/ksl-schema.kdl
zkat: I was thinking this, yes

schema/ksl-schema.kdl
zkat: No, because it's still valuable for reversing definitions when using `ref`. ("I want this node definition, except for this one child")

schema/ksl-schema.kdl
zkat: Schemas are entirely semantic analysis. The syntax is done by the KDL parser itself. This is a second, semantic, pass.

schema/ksl-schema.kdl
zkat: I workshopped a few names (like `multiple`, and `repeatable` ended up feeling the best to me when I was actually writing out a schema

schema/ksl-schema.kdl
zkat: I'm fine with ref. It has good enough precedent.

schema/ksl-schema.kdl
zkat: We can probably unify these, yeah.   I don't like `has-value`, though. Maybe literally just `value`.

schema/ksl-schema.kdl
zkat: There's a convention in KDL about having homogenous child lists simply use `-` as the node name.

schema/ksl-schema.kdl
zkat: the formats correspond to type annotations, but they do not specify the annotation that needs to be present. That is, `format kpath` will validate `"foo"` as if it were `(kpath)"foo"`, but the annotation is not required. To do that, use the `annotation` node.

schema/cargo.kdl
zkat: Thinking through it, the main difference here with JSONSchema is that it allows other properties by default. That is, `"$schema": "foo"` won't fail validation.

It would be a weaker validation, but we could always just default to allowing all children/props/args, and replace `allow-others` with `disallow-others`. Basically, make things more permissive by default, but let people constrain stuff. We can also add an `allow-ksl` to specifically allow KSL-related keywords.

schema/cargo.kdl
zkat: removed

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+        required
+        children {
+            node id about="""
+                Schema identifier
+
+                The unique identifier for this schema. MUST be a valid URL/IRL. Implementations MAY attempt to visit it, but MUST NOT assume it is valid.
+                """ {
+                arg {
+                    type string
+                    format url irl
```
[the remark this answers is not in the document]
zkat: I like `format`.

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                }
+            }
+            node title about="""
+                Schema title
+
+                The title of the schema or the format it describes.
+                """ {
+                ref lang-mixin
```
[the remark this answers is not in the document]
zkat: I've changed things so there's only one way for `ref` to work, and that's `kpath`. 90% of the time, it'll be able to work with just a node name (thus, `ref my-node-mixin`), but if you want to get fancy, that's a thing you can do now, too.

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                An author for the schema.
+                """ {
+                repeatable
+                arg description="Person name" {
+                    optional
+                    type string
+                }
+                prop orcid description="The ORCID of the person" {
+                    type string
+                    pattern #"\d{4}-\d{4}-\d{4}-\d{4}"#
```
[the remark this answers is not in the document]
zkat: I'll revisit this as far as defining custom formats. I still like `pattern`

schema/ksl-schema.kdl
```diff
@@ -0,0 +1,925 @@
...
+                valid.
+                """ {
+                repeatable
+                arg about="Link URL\n\nA URL that the link points to." {
+                    type string
+                    format url irl
+                }
+                // TODO: expand a bit more about what it means for the enum to be `allow-others`.
+                prop rel about="Link relationship\n\nThe relation between the current entity and the URL." {
+                    required
```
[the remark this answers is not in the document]
zkat: I'm game.

schema/ksl-schema.kdl
zkat: I'll remove default nodes. At least for now.

schema/ksl-schema.kdl
zkat: `warn-only` seems reasonable.

schema/ksl-schema.kdl
zkat: It's useful at the schema level, because it communicates what the value will be if no value is provided. It does not inject the value itself, of course--that is the application's responsibility.

And as I mentioned elsewhere: there's nothing about KSL with regards to the parsed AST. It is simply a semantic validation.

schema/cargo.kdl
TheLostLambda: I'll have more time to reply to other comments soon, but for this one, I do like KSL being much more stringent (disallowing others by default), but I would be okay with something like `allow-ksl` if you think it would be useful — though maybe that only makes sense if there are several KSL keywords eventually?

Unfortunately, I don't think that would solve the issue of an application not using KSL, just using `serde` and me wanting to add that LSP-type support (assuming I can't change the `serde` derive myself). Because in that case there is no schema validation being done (via KSL, anyways). For what it's worth, `knus` / `knuffle` by default disallow unknown nodes, so I couldn't get schema-enhanced editor support in any of my current projects unless I could have `/-@ksl:schema` commented.

But I'm happy with what you originally had + letting the LSP pick up the node in comment form as well? If the user wants to use `@ksl:schema` as a real node / version, then I do think they should be required to specify that in their schema — like you originally had it :)

That's my personal preference, anyways — keeping things very explicit and minimizing special cases :)

schema/kdl-schema.kdl
```diff
@@ -0,0 +1,940 @@
+@kdl:schema "https://github.com/kdl-org/kdl/blob/main/examples/kdl-schema.kdl"
```
shimeoki: personally, i think using a schema annotation like this is a no-go.

in user-facing documents, the only data should be the payload itself. optional and metadata things should be kept either outside of the document, or have no value for the actual data.

also we have no guarantees that a user doesn't have a node with a name 

kdl specification already has a perfect example for this scenario: `/- kdl-version` directive. the schema annotations can use just the same:

```kdl
/- kdl-version 2
/- kdl-schema "https://...
my-node data
```

it is only able to appear at the start of the document, but after `kdl-version`, and has the same syntax rules.

[yaml language server does a similar thing.](https://github.com/redhat-developer/yaml-language-server/blob/main/README.md#using-a-modeline)

kdl schemas themselves (written in ksl) can use this syntax, but i don't see a reason for two different set of rules. ksl uses kdl, so this directive can also be applied to the schemas themselves.

SCHEMA-SPEC.md
```diff
@@ -1,347 +1,254 @@
...
+**Schema modification date**
 
-### `version` nodes
+When the schema was modified. If used multiple times, the most recent date will be considered 'latest'.
 
-The version number of this version of the schema.
+- **repeatable**
+- **arg:** Modification date
+  - **type:** `string`
+  - **format:** `date`, `date-time`
```
shimeoki: using repeatable `modified` nodes for schema modifications dates is actually pretty smart. however, why do we need `published` then? order of nodes is important, so in this scenario

```kdl
modified "2026-09-20"
modified "2026-09-21"
```

we can just say that the first modification date is the "published" date. or backwards:

```kdl
published "2026-09-17"
published "2026-09-18"
```

tells that the other publications are the modification dates.

schema/kdl-schema.kdl
```diff
@@ -0,0 +1,940 @@
...
+                                prop min about="""
+                                    Minimum node count
+
+                                    Minimum number of repeated instances of this
+                                    node that must appear in the same scope.
+                                    """ {
+                                    arg {
+                                        gte 0
+                                        type integer
+                                    }
```
shimeoki: probably just an error that a prop uses `arg`. the same applies to the `prop max` below

schema/kdl-schema.kdl
```diff
@@ -0,0 +1,940 @@
...
+            node description about="""
+                Schema description
+
+                A description of the schema or the format it validates, which
+                may include its purposes, its usage, and even examples.
+                """ {
+                arg about="Description text" {
+                    type string
+                }
+            }
```
shimeoki: i think that if a node is used instead of an argument or a property, then it means one of the following:

1. the data cannot be represented as a simple entry (e.g. just a boolean, string, number or null), it is a nested structure.

2. the same data can be repeated multiple times (cannot be a property) and it also should have a name (cannot be represented as arguments).

i think if `title` and `description` are already defined as nodes, then it's worth it to make them repeatable. it can be useful for writing the title or description in different languages. a property like `lang` can be used to optionally tell about the language.

schema/kdl-schema.kdl
```diff
@@ -0,0 +1,940 @@
...
+                                By default, all declared child nodes are
+                                optional. Including this option will require
+                                that this node always appear in its parent's
+                                children block.
+                                """ {
+                                arg {
+                                    type boolean
+                                    default #true
+                                }
+                            }
```
shimeoki: should this be a node, if only one argument is allowed? and that removes the funky thing that "if required is used, then it has a single argument `#true` by default", because the user is just going to write `required=#true`. more symbols, yes. however:

i am going to argue that `required` should be removed as a whole, because `repeatable` exists. otherwise a situation where `required` is true, but also `repeatable max=0` is present.

is `required` shorter than `repeatable min=1`? yes. but i don't think that's marginally worse and this reduces complexity a bit.

shimeoki (changes requested):

schema/kdl-schema.kdl
zkat: Oh I love this

schema/kdl-schema.kdl
zkat: I think aside from whether something is equivalent or longer or shorter, communicating intent with something like this has a lot of value for folks reading it, or for tooling that's trying to express errors

"This thing is required" is a much better error than "At least one of this thing must be present". The latter implies that repetition is expected

SCHEMA-SPEC.md
zkat: I think there's value in being specific about meaning, rather than expecting "modified" to stand for a creation date, much like filesystems have both values (and I'm aware regular file systems don't have the ability to express multiple modified dates like this)

schema/kdl-schema.kdl
zkat: I like this a lot! This is much better than a weird meta node

schema/kdl-schema.kdl
zkat: oh yeah idk what happened here.

SCHEMA-SPEC.md
shimeoki: yes, filesystems don't allow multiple modified dates - only the last change. therefore "published" and "modified" are separate, because they are essentially key-value pairs.

nevertheless, fair enough. multiple "modified" dates are probably going to be confusing. i think "published" is better, but if you want to separate them, i won't object.

however, maybe something else for versioning can be done then? multiple "modified" or "published" dates are a cool idea, but how it can be used in the end? maybe allow a label of some kind? or a url?

```kdl
published "2026-09-17" title=v1
published "2926-09-18" title=v2-draft
```

schema/kdl-schema.kdl
shimeoki: i can argue that a language server (or just a tool) that implements verification/validation of the schema can represent an error of `repeatable min=1` to the user as "this node is required at least once", which is clear enough. and this may be easier then having two kinds of errors: "this node is required" for `required` and "this node should appear at least X times and at most Y times" for `repeatable`.

while schemas may be read by people (and is written by a person), it's probably mainly going to be used as an outline for a program to then verify the documents automatically. for that, i think reducing the amount of rules used and possibilities of invalid schemas in the first place (e.g `required` and `repeatable max=0`) is able to simplify implementations.

SCHEMA-SPEC.md
zkat: for the sake of shaving the yak a little: what about `label` as a generic name for the identifier? Just because `title` already means something in ksl

```kdl
published "2026-09-17" label=v1
published "2926-09-18" label=v2-draft
```

schema/kdl-schema.kdl
zkat: I'm still kind of on the fence about this, tbh. I think it would be good to have others chime in about it and see.

SCHEMA-SPEC.md
shimeoki: oh, yeah, that's better. i have written "label" in the comment and "title" in the block, idk why

PR #487, opened 2025-01-18
Fix typo: RFC1123 is the RFC that defines requirements for hostnames

clarfonthey:
RFC1123 is [Requirements for Internet Hosts -- Application and Support](https://www.rfc-editor.org/rfc/rfc1123)
RFC1132 is [A Standard for the Transmission of 802.2 Packets over IPX Networks](https://www.rfc-editor.org/rfc/rfc1132)

This change applies to V1, V2, and the schema spec.


zkat (approved):

zkat:
oops. Thanks!

PR #488, opened 2025-01-18
Add the html standard, in kdl format, as a large-document benchmark.

tabatkins:
Per #482, starting a benchmarks collection. First up, the HTML standard, which isn't particularly deep but *is* nearly a million lines long.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -54,3 +54,7 @@ please send a PR.
 If you think the disagreement is due to a genuine error or oversight in the
 KDL specification, please open an issue explaining the matter and the change
 will be considered for the next version of the KDL spec.
+
+# Benchmarks
+
+The `benchmarks` folder contains some large or gnarly documents intended to be used to stress-test your parser and help with profiling. They are intentionally not part of the testsuite, and just provided for your own personal benefit.


zkat:
Incredible. 20mb. Lmao

how big is the original, out of curiosity?

bgotink:
The HTML text is 13.8 MB, but I'd like to note that the KDL version includes a lot more whitespace because the HTML whitespace is handled just like HTML text (i.e. included as `-` nodes) and the KDL document itself is indented.

Example:

```html
<p>
  This is
  a paragraph
</p>
```

is currently encoded as

```kdl
p {
	- "\n  This is\n  a paragraph\n"
}
```

I've run a cleanup step over the KDL text to turn this kind of node into

```kdl
p " This is a paragraph "
```

and then [the resulting KDL document](https://github.com/user-attachments/files/18469425/whatwg-html.kdl.txt) is 16.1 MB.

zkat:
I was about to ask whether you’d like to PR the cleaned up version but tbh I think the huge one is gonna be better as a benchmark so :)

tabatkins:
I went ahead and added the compacted version *alongside* the larger version. Having a document with similar complexity but 3/4 the size can be useful for checking traces for differences.

PR #499, opened 2025-01-23
Remove mandatory node-space before slashdash and children block

### LINKED ISSUE #495: Node space now required before children block
Between v1 and v2, the grammar changed, requiring `node-space` before `node-children`:

```
base-node := slashdash? type? node-space* string
      (node-space+ slashdash? node-prop-or-arg)*
      // slashdashed node-children must always be after props and args.
      (node-space+ slashdash node-children)*
      (node-space+ node-children)?
      (node-space+ slashdash node-children)*
      node-space*
```

What I don't remember, was this change intentional (I don't see it in the changelog), or was it not? If not, is it a blessing in disguise or should we revert?

larsgw:
Fix #495


tabatkins:
Could you drop a test for this?

zkat (approved):
A couple of interesting tests will be good yeah

larsgw:
Okay, added.

bgotink:
Test case `braces_in_bare_id_fail.kdl` becomes valid with this change so it should be updated as well.

It's content

```kdl
foo123{bar}
```

is now identical to

```kdl
foo123 { bar }
```

larsgw:
Indeed, I forgot to patch that into kdljs while testing. Thank you @bgotink

tests/test_cases/expected_kdl/braces_in_bare_id.kdl
bgotink, line 1: Could we change the file contents to

```kdl
foo123 {
    bar
}
```

to align with the other files in `expected_kdl`?

I use the tests to not only validate my parser but also my formatter, and inconsistencies lead to problems on my end (see also #501)

larsgw: Of course, I must have changed that in the wrong directory.

bgotink (approved):

PR #506, opened 2025-02-18
Escape U+200E and U+200F

### LINKED ISSUE #502: Newly added `html-standard-compact.kdl` file is not valid
There are a couple places in this file that use literal UTF-8 code points that, I believe, are disallowed in KDL:

1. On line `491688` there is a literal `U+200E` character.
2. On line `497376` there is a literal `U+200F` character.

bgotink:
Fixes #502

Caused by improper escaping, solved in https://github.com/bgotink/kdl/commit/d91d5821c506ddd86e6b378e3d344ffedb6c2d93. Turns out "JSON.stringify should get us most of the way there" was not a great thought process for the formatter 😅


zkat:
is the other file not affected, btw?

bgotink:
The other file was generated using a small script (https://github.com/kdl-org/kdl/discussions/482#discussioncomment-11874804) which properly escaped the U+200E and U+200F in the text.

PR #509, opened 2025-02-24
rename test case to better describe error

alightgoesout:
After removing the requirements to have spaces before node children (#499), I feel like test cases `brackets_in_bare_id_fail.kdl` is poorly named.

```kdl
foo123{bar}foo weeee
```

This test should still fail, but for different reasons. The name suggest `foo123{bar}foo` is not a valid identifier, but it should now be parsed as `foo123 { bar }`. It fails because there are no node separator between the first node and the second one (`foo weeee`).

I have looked and found no test that check a similar case, so I have renamed the test case to `semicolon_after_children_fail.kdl`, but feel free to suggest other names.


tabatkins:
This sounds completely reasonable, but I think the name still sounds misleading. ^_^ Perhaps `semicolon_missing_after_children_fail`?

alightgoesout:
Yes, you're right. I have renamed the file accordingly.

Issue #510, opened 2025-02-24
Idea: number suffixes as annotations


zkat:
There was a discussion over at https://github.com/YaLTeR/niri/discussions/1142#discussioncomment-12287124 when trying to come up with enhancements to Niri's DSL regarding numerical stuff. The tl;dr is that there's interest in being able to type things like `100px` and `50%`, and not having to quote it.

I thought about this a bit, and I would like to propose the following change to the *number* grammar to facilitate this. This change would be forward compatible: existing v2 parsers would error (instead of returning incorrect data).

The change is as follows:

1. any valid number may be suffixed, with no spaces, by a string (of any kind, but most likely an unquoted string). If this is done, the suffix will be used as the *type annotation* for that number value. That is, `100px` would be equivalent to `(px)100`.
2. If you have BOTH a parenthetical annotation, AND a suffix, it will be treated as a syntax error. You simply cannot do `(px)100%`

One catch: a very narrow range of invalid v2 documents will now become valid (since `foo 100px` will no longer be an error)

Thoughts?

tabatkins:
I think it's an incredible idea, but as a CSS author, I'm a little biased. ^_^

Seriously, tho, I think it really is a very good idea. It's *often* the case that you don't need to annotate your numbers, but there are plenty of cases where numbers are *heavily* annotated, and this would be a huge boon to those. I think the wide-spread basic familiarity with CSS makes this a pretty natural thing to read and understand (and *useful* for CSS interop), and while you have to make the leap to "the unit is the tag", that's also really the *only* logical interpretation for it.

> One catch: a very narrow range of invalid v2 documents will now become valid (since foo 100px will no longer be an error)

This is true of literally any change that's not strictly tightening the parsing; it's perfectly fine unless we have reason to believe that's a common error. KDL's lack of error recovery makes this more or less a non-issue, imo.

bgotink:
I assume this wouldn't apply to hex numbers? I don't think anyone wants `0xffpx` to be allowed.

tabatkins:
I assume so, yeah. Just decimals.

zkat:
@tabatkins i was hoping you would chime in! I had a feeling it would be either "perfect, no notes" or " beware! Run! Here be dragons!"

I'm still curious whether you think there's any gotchas about this based on your experience, but it doesn't sound like it.

And yeah: originally I was gonna say these should parse as strings, but having them be numbers-with-special-type-syntax just clicked the moment I thought of it.

[...]

PR #513, opened 2025-03-30
number suffix type annotations

zkat:
Fixes: https://github.com/kdl-org/kdl/issues/510


zkat:
@tabatkins honestly the most heartbreaking thing about all this to me is that I can't have `123u32` and such :( Gotta do `123#u32`

draft-marchan-kdl2.md
```diff
@@ -265,10 +265,65 @@ KDL does not specify any restrictions on what implementations might do with
...
+#### Explicit Suffix Type Annotation
+
+Any ({{number}}) may have a `#` attached to it, followed by any valid
+({{identifier-string}}). This is an explicit ({{suffix-type-annotation}}) syntax
+without any of the relatively complex requirements of
+({{bare-suffix-type-annotation}}), which can be a useful escape hatch. For
+example: `10.0#u8` is invalid syntax without the `#` prefix.
+
+Note again that, unlike Bare Suffixes, Explicit Suffixes may be used with ALL
+({{number}}) formats (hexadecimal, decimal, octal, and binary).
```
tabatkins, line 321: I might drop an example here.

```suggestion
({{number}}) formats (hexadecimal, decimal, octal, and binary).
For example, `0x1234#u16` is valid.
```

zkat:
@bgotink does this align with your understanding/experience/what you did in your lib?

zkat:
alright, tests added. This is ready for final review.

draft-marchan-kdl2.md
```diff
@@ -265,10 +265,66 @@ KDL does not specify any restrictions on what implementations might do with
...
 
-Additionally, the following type annotations MAY be recognized by KDL parsers
-and, if used, SHOULD interpret these types as follows:
+### Suffix Type Annotation
+
+When a ({{value}}) is a ({{number}}), it's possible to attach the type
+annotation as a "suffix", instead of prepending it between `(` and `)`. This
+makes it possible to, for example, write `10px`, `10.5%`, `512GiB`, etc., which
+are equivalent to `(px)10`, `(%)5`, and `(GiB)512`, respectively.
+
```
clarfonthey, line 274: Just for readability purposes, I think it's worth mentioning the `#` escape hatch here with a reference to the "Explicit Suffix Type Annotation" section, at least when it comes to types like `u32`. For example, maybe:

> To remove ambiguity, some suffixes must be prefixed with `#`: for example, `10.0u8` is invalid, but `10.0#u8` is. The full list of rules for invalid suffixes is clarified in the "Explicit Suffix Type Annotation" section.

tabatkins (changes requested):
r+ after review of suggested changes

tests/test_cases/input/suffix_type_bare_underscore_fail.kdl
bgotink, line 1: `integer` can end on an underscore so this is actually valid and equivalent to `(abc)123_`

bgotink (changes requested):
This matches what I implemented apart from one test that appears to be wrong and one mistake on my end where the parser skips certain validations on `#` suffixes which makes `123#123` equivalent to `("123")123` which is wrong.

tests/test_cases/input/suffix_type_bare_underscore_fail.kdl
tabatkins: correct. this is why we can't *start* a bare suffix with `_`, becuase the syntax would parse differently than intended.

draft-marchan-kdl2.md
clarfonthey: Ah, GitHub collapsed it as "outdated" because they're bad

bgotink:
The following tests that previously failed now run successfully:

| test name | test | equivalent document |
|-----------|-----|----------------------|
| `bare_ident_numeric_fail.kdl` | `node 0n` | `node (n)0` |
| `bare_ident_numeric_sign_fail.kdl` | `node +0n` | `node (n)+0` |
| `illegal_char_in_binary_fail.kdl` | `node 0bx01` | `node (bx01)0` |
| `multiple_x_in_hex_fail.kdl` | `node 0xx10` | `node (xx10)0` |
| `no_digits_in_hex_fail.kdl` | `node 0x` | `node (x)0` |

zkat:
Uggghhhhh. That makes sense. Looks like we’re gonna need to be more specific what order things run in. I have an idea for the grammar.

zkat:
@bgotink 
> The following tests that previously failed now run successfully:

Looking at this...

These should be dropped because they're ok now:

> test name 	test 	equivalent document
> `bare_ident_numeric_fail.kdl` 	`node 0n` 	`node (n)0`
> `bare_ident_numeric_sign_fail.kdl` 	`node +0n` 	`node (n)+0`

These tests should stay and continue to fail. Once you've hit a `0b`/`0o`/`0x` number prefix, you SHOULD only be able to parse their related number formats:

> test name 	test 	equivalent document
> `illegal_char_in_binary_fail.kdl` 	`node 0bx01` 	`node (bx01)0`
> `multiple_x_in_hex_fail.kdl` 	`node 0xx10` 	`node (xx10)0`
> `no_digits_in_hex_fail.kdl` 	`node 0x` 	`node (x)0`

zkat:
uggghhhh I understand why those last 3 tests pass now. I want to do something about it, though. I don't like that. I wonder if there's a good way around it.

zkat:
@bgotink @tabatkins I've... kind of changed the rules a bit. They're simpler now. And most importantly: we have `0u64` available now!

I checked #510 and I didn't see any discussion about us tackling these rules, because we were focusing so much on what the rules for the _suffix_ should be that we didn't take a step back and think of these numbers as a whole, and how the number syntax in general could be addressed.

Please lmk what you think. I think with these changes, we'll get the results I expected in https://github.com/kdl-org/kdl/pull/513#issuecomment-2813782924

zkat:
Some key changes:

* I reorganized things a bit in general.
* The grammar itself now blocks simultaneous suffix and prefix annotations
* The underscore `_` initial character restriction was removed: a bare suffix CAN'T start with one because the integer will slurp it up first
* The complex rules meant to disambiguate from non-decimals were removed. The fact that we only allow bare prefixes on decimals is sufficient to protect us here imo. I think these complex rules formed because things were moving fast and we didn't take a step back and really rethink the implications of only doing decimals. This has been clarified in the spec. With the new rules, there is no dangerous ambiguity, just potential syntax errors on certain zero values, which is notably NOT an unexpected parsing success, which I think is the dangerous bit.
* The rules around exponential-likes have been changed a bit to guard against small typos if you miss the digit part (so `1e+` is illegal)

zkat:
I'm also wondering: could we just drop the exponent restriction (but keep the `e+` protection, in case someone fails to write that digit and ends up with an unfortunate parse)

zkat:
TODO: I'm gonna try changing the explicit suffix syntax to use parentheses. I'm not really sure why we wouldn't just do that.

mwh:
I don't think the specification prose as currently written does require or even allow a trailing `_` to be consumed by the number ("digits ... may be *separated* by `_`"), but the authoritative grammar does include arbitrary underscores in place of digits anywhere except the beginning.

I think either the description of the grammatical language needs tightening, or the grammar may now allow some undesirable constructions with underscores and suffixes. Specifically, I'm not sure whether `*` is expected to commit a parser to whatever it finds the first time, or whether it can backtrack to allow later productions to match. Here, it would be backtracking to an underscore after the input didn't match when the underscore was consumed as part of a number.

Consider `12_3,x`. The overall match would fail when `(digit | '_')*` from the integer production consumes `_3`; does it then backtrack and try consuming less of the input? If it stops short after "12" and leaves "_3,x", the whole input matches suffixed-decimal successfully with significand accepting "12" and bare-type-suffix accepting "_3,x" — so the result would be equivalent to `(_3,x)12`, though I think it should be an error. `12_3.4` and `12_3.4.5` differ at the same point, or consider `1_234,567`. All of these can be produced *from* the grammar, at least.

I'm not sure whether I am misreading the description of the grammar language. `*` consumes "as many instances as possible without failing the match" — is that failing the match of the whole input, or does it just mean as many instances as are present at this point in the input and failure isn't really part of it? The comparison to standard regex semantics and the existence of cut points makes me think it can shorten if needed. If it does commit early to the longest sequence it finds, this issue doesn't come up. Otherwise, for integer to definitively slurp the underscore up and make this an error, I think there would need to be a cut point `suffixed-decimal := significand ¶ (bare-type-suffix | (exponent? explicit-type-suffix))`.

If there is an issue, either banning `_` as the initial character of a suffix again or solidifying the grammatical handling of integers could address it. I am not personally a fan of baking parsing and backtracking rules into a grammar and would probably just block `_`, but there are reasons to have that kind of grammar too, particularly to constrain where an error is detected.

Something shaped like `1_234,567` seems like the primary case where this is realistic and actually matters, or someone trying to write a list `1_234, 5_678, ...`. I do think the grammar should rule these out, but if I were implementing a parser directly off this grammar right now, I would end up with this backtracking and unknowingly accepting these cases because nothing commits the parse to the path that produces an error. If I built the parser with a lexer in the front, I think the lexer would probably consume the whole number as expected and then I wouldn't encounter the issue. Clearly one of those is wrong, but it's not good for compatibility if both seem reasonable to make.

If nothing else, test cases including both underscores within the number and invalid suffixes will be good to ensure that the incorrect readings are detected.

---

I also wonder in a similar way about `0xaz`. The prose does rule this out. This time it's the grammar semantics of `-` in significand-initial I'm not certain about: "any `digit` except something that matches the literal `'0x'`" seems like it'd be the same as just digit alone, because "0x" is not a digit and is not matched by digit. The intention here is clear, just the formalism may not match up.

draft-marchan-kdl2.md
```diff
@@ -1013,12 +1102,27 @@ multi-line-raw-string-body :=
...
+    bare-type-suffix
+    | (exponent? explicit-type-suffix)
+)
+significand := sign? significand-initial integer? ('.' integer)?
 exponent := ('e' | 'E') sign? integer
 integer := digit (digit | '_')*
+significand-initial = digit
+    - '0b'
+    - '0o'
+    - '0x'
```
tabatkins, line 1116: This use of `-` isn't quite right - it's not subtracting from the LHS (the `digit`), it's imposing a wider restriction on the digit *and* its following character (which in this grammar would come from the start of `bare-type-suffix`).

I think this should instead be handled as a cut point on the `hex`/`octal`/`binary` productions, so that if you see their prefix you're committed. (We'll also have to specify that the ordering of `|` alternates *is* significant when cuts are involved, so `normal-value` is parsed first and gets to claim that syntax space with a cut before `suffixed-decimal` can run and claim it.)

draft-marchan-kdl2.md
```diff
@@ -1013,12 +1102,27 @@ multi-line-raw-string-body :=
...
 digit := [0-9]
 sign := '+' | '-'
 
+bare-type-suffix := bare-type-suffix-initial identifier-char*
+bare-type-suffix-initial := identifier-char
+    - '.' - ','
+    - (('e' | 'E') (sign | digit))
+explicit-type-suffix := '#' identifier-string
```
tabatkins, line 1124: Similarly, this isn't a valid use of `-`, since it's not subtracting from `identifier-char`. Easier to fix, tho:

```suggestion
bare-type-suffix := (bare-type-suffix-initial identifier-char*) | 'e' | 'E'
bare-type-suffix-initial := identifier-char - ('.' | ',' | 'e' | 'E') | identifier-char{2} - (('e' | 'E') (sign | digit))
    - '.' - ','
    - (('e' | 'E') (sign | digit))
explicit-type-suffix := '#' identifier-string
```

This allows `1e` and `1egg`, but disallows `1e+`, `1e1`, and `1e+gg`.

tabatkins (changes requested):

tabatkins:
> Consider 12_3,x [snip discussion about why this successfully matches the current grammar]

Yup, the `_` prefix restriction was actually load-bearing unless we introduce more cut-points, which I'd like to avoid except when necessary. (They're easy to forget about and then accidentally have side-effects.) We need to bring it back for the grammar (even if it might not be worth calling out in the prose, tho I actually *do* think some discussion about prefixes that won't work as expected due to parsing rules is worthwhile in the prose).

tabatkins:
>  I think these complex rules formed because things were moving fast and we didn't take a step back and really rethink the implications of only doing decimals. This has been clarified in the spec. With the new rules, there is no dangerous ambiguity, just potential syntax errors on certain zero values, which is notably NOT an unexpected parsing success, which I think is the dangerous bit.

This isn't correct, there's still an unexpected parsing success in some cases. See `20x1`, for example - totally valid under these rules (as expected), equivalent to `(x1)20`. But if the value becomes 0, it misparses - `0x1` parses equivalent to `1`, not `(x1)0`. That's the scenario we were afraid of, and the reason those restrictions existed.

If the part following the `x` *doesn't* look like hex digits (`20xfoo`), then sure, when the value is 0 it'll just be a parse failure rather than a misparse. (We *were* somewhat worried about that as a usability issue, but I agree it can be a reasonable bullet to bite.) But the misparse when the following *does* look like hex digits is the real problem.

draft-marchan-kdl2.md
tabatkins: Dang it, this doens't work, the second clause in `bare-type-suffix-initial` lets `.f` smuggle in as valid. I'll need to tweak further.

tabatkins: Okay, fixed, using the same techniques we use in idents to avoid the ambiguous numeric patterns.

PR #515, opened 2025-04-27
Fix expected value for `multiline_string_whitespace_only` test

nike4613:
The spec states:

> A Multi-Line String is opened and closed by three double-quote characters, like """. Its first line MUST immediately start with a Newline ({{newline}}) after its opening """. Its final line MUST contain only whitespace before the closing """. All in-between lines that contain non-newline, non-whitespace characters MUST start with at least the exact same whitespace as the final line (precisely matching codepoints, not merely counting characters or "size"); they may contain additional whitespace following this prefix. The lines in between may contain unescaped " (but no unescaped """ as this would close the string).
> 
> The value of the Multi-Line String omits the first and last Newline, the Whitespace of the last line, and the matching Whitespace prefix on all intermediate lines. The first and last Newline can be the same character (that is, empty multi-line strings are legal).

This does NOT state that whitespace-only lines are omitted from the value. The test already acknowledges this; the final line in the third string contains only 4 spaces. By this logic, the first line should contain 3 spaces, based on the input file.

The spec is unclear what to do with the second line. It permits whitespace-only lines to not have the required prefix, but does not specify whether such a line is included in the value. This test implies that it should not be, and this change keeps that unchanged.


tabatkins:
The spec itself isn't super clear on this point - it's only illustrated in an example currently (<https://kdl.dev/spec/#section-3.12.2.3>). But the test is correct. All whitespace-only lines are treated as having no value, just an empty line in the string. It is definitely not intended that a whitespace-only line that *happens* have the correct prefix has the prefix removed and represents the rest of its line in the string value, while a whitespace-only line that doesn't have the correct prefix has its value ignored entirely; that would be a very fragile definition.

tabatkins:
(Spec was fixed to be state this normatively in #516, btw)

PR #538, opened 2025-07-14
Kdl4J now supports KDL v2

alightgoesout:
With the release of Kdl4J 1.0.0, it now supports KDL v2.


zkat:
yay! Make sure to send a PR to the site, too!!

PR #552, opened 2026-01-28
Fix multi-line-string-body grammar

brandonchinn178:
`multi-line-raw-string-body` is already non-greedy:
```
multi-line-raw-string-body :=
    (unicode - disallowed-literal-code-points)*?
```
The grammar seems incorrect if it's greedy, since it'll always consume the newline following it:
```
quoted-string :=
    '"' single-line-string-body '"' |
    '"""' newline
    (multi-line-string-body newline)?
    (unicode-space | ws-escape)* '"""'
```
Furthermore, the grammar currently always requires `string-character` after `"` or `""`, which doesn't allow this value:
```kdl
"""
""
"""
```


zkat:
Could you add a couple of tests for these cases to our compliance suite?

brandonchinn178:
@zkat I added test cases for the multiline-string-with-only-quote-content repro. The other ones should be covered by normal multiline string checks.

Also, in case you didn't see it, it'd be really great to pass my compliance validator over to you, superceding the current compliance suite: https://github.com/kdl-org/kdl/issues/553

PR #554, opened 2026-02-10
Add Zig support to README compatibility table

desttinghim:


zkat:
Could you add this to the website repo too?

PR #560, opened 2026-04-16
Update README.md

njreid:
Added Golang library that supports both v1 and v2, and passes kdl-rs and kdl-test.


njreid:
Thanks @tabatkins, have a great weekend.

PR #566, opened 2026-08-31
add kdly to Go implementations

shimeoki:
while it's still in alpha, i want to share this project and include it in the implementations list. also, as a "proof of concept" that it works, i built [a formatter](https://codeberg.org/shimeoki/kdlf) on top of it.

and i wanted to ask a question about tooling. for example, there are [kdlfmt](https://github.com/hougesen/kdlfmt), [kdl-fmt](https://github.com/dj95/kdl-fmt), [kdl-lsp](https://github.com/kdl-org/kdl-rs/tree/main/tools/kdl-lsp) from [kdl-rs](https://github.com/kdl-org/kdl-rs) and my aforementioned formatter. should there also be a separate section for tools? while implementations are more important, i think keeping track of projects to interact with the language is still worth mentioning.


zkat:
Having a tools section sounds like a great idea!

Incidentally: Please make this PR against our website repo as well!

